PR #73, opened 2019-08-13
fix Dialog in Edge browser

YingXue:
Chrome is more tolerent when a property value of a componenet is undefined. 
Disable rendering ComplexReportedFormPanel to prevent it.


src/app/devices/deviceContent/components/deviceSettings/deviceSettingsPerInterfacePerSetting.tsx
```diff
@@ -168,13 +167,15 @@ export default class DeviceSettingsPerInterfacePerSetting
         const { reportedTwin, settingModelDefinition : modelDefinition, settingSchema : schema } = this.props;
         return (
             <div role="dialog">
```
montgomp: I think it would be better to move the `{this.state.showReportedValuePanel &&` up one line to the `<div role="dialog">` so that the entire component is on or off. You might need to include a `||<></>`.

PR #79, opened 2019-08-16
Helper tests and fixes

YingXue:
Added some util tests
Removed some duplicate code
Added a workaround for the test console warning


package.json
```diff
@@ -154,13 +154,15 @@
     "collectCoverageFrom": [
       "src/**/*.{ts,tsx}",
       "!src/**/*Container.{ts,tsx}",
+      "!src/app/devices/saga.ts",
```
montgomp: I'd rather not skip adding tests for this.

YingXue: This is the root saga, so there isn't anything to test here:

export default function* rootSaga() {
    yield all([
        ...deviceListSagas,
        ...deviceContentSagas,
        ...notificationsSagas,
    ]);
}

YingXue: Added it back, since we do want to test the root saga. Going to really test it in a different PR afterwards. @montgomp

PR #81, opened 2019-08-19
Enable submit on enter on Add Device view.

andrew-buckley:


package.json
```diff
@@ -57,14 +57,14 @@
   },
   "repository": {
     "type": "git",
-    "url": "git+https://github.com/Azure/azure-iot-pnp-template-management.git"
+    "url": "git+https://github.com/Azure/azure-iot-explorer.git"
```
YingXue: good catch!

PR #83, opened 2019-08-19
localize displayname and displayunit in contents

YingXue:
Display localize text for displayname and displayunit for commands, props and events
Fix accessibility bug where certain aria role expect child items. Use a different role when there is not child


src/app/devices/deviceContent/components/deviceEvents/deviceEvents.tsx
```diff
@@ -199,7 +199,7 @@ export default class DeviceEventsComponent extends React.Component<DeviceEventsD
                 loadMore={this.fetchData}
                 hasMore={hasMore}
                 loader={this.renderLoader(context)}
-                role="feed"
+                role={this.state.events && this.state.events.length === 0 ? '' : 'feed'}
```
andrew-buckley: Can role be empty?

YingXue: The infinite scroll handles this case fortunately :)

PR #96, opened 2019-08-22
Adds a node build and test workflow/action.

andrew-buckley:
This allows us to use GitHub Actions to build and test across our 3 supported operating systems using node versions 8, 10, and 12


src/localization/resourceKeys.ts
```diff
@@ -1,472 +1,472 @@
...
-      array : {
-         add : "template.array.add",
-         remove : "template.array.remove",
-      },
-      collapse : {
-         close : "template.collapse.close",
-         open : "template.collapse.open",
-      },
-   };
+export class ResourceKeys {
```
YingXue: Anything got change in this file?

andrew-buckley: I don't think so, I think when I build on linux the encoding or line endings change somehow

Issue #97, opened 2019-08-23
[BUG] Failed to retrieve device list


SquintEastwood:
**Describe the bug**
After copying iothubowner (or any other policy's) connection string primary key from Azure and pasting into the IoT hub connection string box, receive this error as a popup:
"Failed to retrieve device list: 
SyntaxError: Unexpected token N
in JSON at position 4" 

**To Reproduce**
1. Go to Iot Hub in Azure Portal
2. Click on an IoT Hub service
3. Scroll down to Shared access policies under settings
4. Click on iothubowner 
5. Copy connection string--primary key
6. Paste key into Azure IoT explorer, either under Settings and app configuration or under App configurations on program startup
7. Click connect

**Expected behavior**
Expected it to connect to Azure and display devices connected to IoT Hub

**Screenshots**
![image](https://user-images.githubusercontent.com/50408735/63599034-bc7f5480-c58e-11e9-8927-f544578299b6.png)

**Desktop (please complete the following information):**
 - OS: Windows 10 Enterprise
 - Build 15063
 - App version 0.10.1

YingXue:
Hi @SquintEastwood,
I will close the issue as 'no repro' for now. Please feel free to repo it if it is still happening.

Thank you.

andrew-buckley:
@SquintEastwood we made a change so that you can open developer tools to get a better idea of what might have been failing. If this is still happening for you it would be helpful to show us the network trace and console log for this error.

PR #102, opened 2019-09-04
Devtools and improvements

andrew-buckley:
Adds support for opening dev tools for debugging the desktop application. Improves bundle size. Mov
es webpack config and electron config to ts and splits webpack into dev and prod.


public/electron.ts
```diff
@@ -66,12 +73,26 @@ function createMenu() {
                 { role: 'minimize' },
                 { role: 'close' }
             ]
+        },
+        {
+            label: 'Help',
+            submenu: [
+                {
+                    accelerator: 'CommandOrControl+Shift+I',
```
YingXue: is it the default shortcut that people follow for electron apps?

src/app/devices/deviceContent/components/deviceContentNav.tsx
```diff
@@ -26,7 +26,7 @@ interface DeviceContentNavState {
 }
 
 export const NAV_LINK_ITEMS_PNP = ['interfaces', 'settings', 'properties', 'commands', 'events'];
-export const NAV_LINK_ITEMS_NONPNP = ['identity', 'twin', 'events'];
+export const NAV_LINK_ITEMS_NONPNP = ['identity', 'twin', 'events', 'methods'];
```
YingXue: snapshot test of this component needs to be updated accordingly. But before we do so, I'd like to add the labels at least on the deviceMethod before we enable it in the route as user would see it in the next release if we enable it

andrew-buckley:
> Thank you so much for making all these wonder changes. I'd like to learn how the devtools works tomorrow from you.

Sure, stop by anytime.

public/electron.ts
andrew-buckley: It's the shortcut that VSCode uses.

src/app/devices/deviceContent/components/deviceContentNav.tsx
andrew-buckley: Yeah I'm fixing the tests now and going to revert the methods change. That shouldn't go in this PR.

andrew-buckley:
#97

PR #103, opened 2019-09-04
finish all saga tests

YingXue:
Added tests for the rest of sagas (mostly digital twin related)


src/app/devices/deviceContent/sagas/digitalTwinInterfacePropertySaga.ts
```diff
@@ -67,7 +67,8 @@ export function* patchDigitalTwinInterfacePropertiesSaga(action: Action<PatchDig
     }
 }
 
-export const generateInterfacePropertiesPayload = (interfaceName: string, propertyKey: string, patchData: object) => {
+// tslint:disable-next-line:no-any
+export const generateInterfacePropertiesPayload = (interfaceName: string, propertyKey: string, patchData: any) => {
```
andrew-buckley: Is this type changed only for the test? Could we supply an object for testing instead of a number?

YingXue: It is more like a bug discovered by the test. The value can be anything not just an object :)

PR #104, opened 2019-09-06
add api tests; small fixes and cleanup

YingXue:
Added all api tests to boost coverage to 80%
Some css cleanup and fix


src/app/api/services/devicesService.ts
```diff
@@ -101,17 +101,13 @@ const dataPlaneResponseHelper = async (response: Response) => {
...
     // error case
-    if (!result) {
-        throw new Error();
-    }
-    if (result.ExceptionMessage && result.Message) {
-        throw new Error(`${result.Message}: ${result.ExceptionMessage}`);
-    }
-    else if (!!result.ExceptionMessage || result.Message) {
-        throw new Error(!!result.ExceptionMessage ? result.ExceptionMessage : result.Message);
+    if (result && result.body) {
```
andrew-buckley: Wondering if this will work as is? I didn't see a corresponding change in server.ts to put this info in the body.

YingXue: Data plane related handling is in a file called dataPlaneHelper.ts. (line 40 to 60)
Everything has been put into the body without this file being changed accordingly, so it is a regresion bug I am trying to fix here.
In addition, we got the feedback that the message is too long, so I am going to only show "Message" here.

PR #108, opened 2019-09-10
add confirmation box for settings pane

YingXue:
added confirmation box for settings pane and also added some aria label


src/app/settings/components/settingsPane.tsx
```diff
@@ -56,80 +58,96 @@ export default class SettingsPane extends React.Component<SettingsPaneProps & Se
...
+                            <span className="helptext">{context.t(ResourceKeys.settings.modelDefinitions.helpText)}</span>
+                            <RepositoryLocationList
+                                items={this.state.repositoryLocations}
+                                onAddListItem={this.onAddListItem}
+                                onMoveItem={this.onMoveRepositoryLocation}
+                                onPrivateRepositoryConnectionStringChanged={this.onPrivateRepositoryConnectionStringChanged}
+                                onRemoveListItem={this.onRemoveListItem}
+                            />
+                        </section>
+                        {this.renderConfirmationDialog(context)}
```
YingXue: This line is the change in this function. Everything else is spacing.

PR #118, opened 2019-09-19
add route constants; enable direct method page

YingXue:
image


src/localization/locales/en.json
```diff
@@ -100,7 +100,7 @@
             "interfaces": "Interface",
             "events": "Telemetry",
             "eventsPerInterface": "Telemetry",
-            "methods": "Device Methods",
+            "methods": "Direct Method",
```
andrew-buckley: Wonder if this should be changed to "Direct methods" to follow the capitalization of the other menu items.

andrew-buckley: Please check elsewhere too

src/app/devices/deviceContent/components/deviceMethods/deviceMethods.tsx
```diff
@@ -45,119 +44,133 @@ export default class DeviceMethods extends React.Component<DeviceMethodsProps &
...
 
+    private readonly showCommandBar = (context: LocalizationContextInterface) => {
+        return (
+            <CommandBar
+                className="command"
+                items={[
+                    {
+                        ariaLabel: context.t(ResourceKeys.deviceMethods.invokeMethodButtonText),
+                        disabled: !this.formReady(),
+                        iconProps: {iconName: REFRESH},
```
andrew-buckley: Is there a better icon for this?

src/localization/locales/en.json
YingXue: Good catch! I was only paying attention to the portal...

src/app/devices/deviceContent/components/deviceMethods/__snapshots__/deviceMethods.spec.tsx.snap
```diff
@@ -0,0 +1,95 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`deviceMethods matches snapshot 1`] = `
+<Fragment>
+  <StyledCommandBarBase
+    className="command"
+    items={
+      Array [
+        Object {
+          "ariaLabel": "deviceMethods.invokeMethodButtonText",
```
andrew-buckley: Should this change to `directMethods` to match the changes elsewhere?

andrew-buckley:
I see a few discrepancies in the usage of direct vs device method throughout the code. I'd just have a scan through and see where things should be changed. I also wonder if it should be `Direct methods` rather than `Direct method` to match the portal.

src/app/devices/deviceContent/components/deviceMethods/deviceMethods.tsx
YingXue: Maybe we can use what portal use

YingXue:
> I see a few discrepancies in the usage of direct vs device method throughout the code. I'd just have a scan through and see where things should be changed. I also wonder if it should be `Direct methods` rather than `Direct method` to match the portal.

Portal uses 'Direct Method'. So I will be using 'Direct method' here to match this tool's convention.

PR #126, opened 2019-10-09
Create a dark theme, they said. It'll be fun, they said.

montgomp:


asrudra:
use React Context with hook API to handle theme change

src/app/devices/deviceContent/components/directMethod/__snapshots__/directMethod.spec.tsx.snap
```diff
@@ -61,6 +61,7 @@ exports[`directMethod matches snapshot 1`] = `
                 "automaticLayout": true,
               }
             }
+            theme="vs-dark"
```
YingXue: I find it mysterious that the last commit which made light theme as default, these snap shot are still passing

src/app/settings/components/settingsPane.tsx
```diff
@@ -128,13 +132,26 @@ export default class SettingsPane extends React.Component<SettingsPaneProps & Se
...
+                        </section>
                         {this.renderConfirmationDialog(context)}
                     </Panel>
                 )}
             </LocalizationContextConsumer>
         );
     }
 
+    private readonly setTheme = (ev: React.MouseEvent<HTMLElement>, darkTheme: boolean) => {
+        this.props.onSetTheme(darkTheme ? Theme.dark : Theme.light);
```
YingXue: It's my thought that we could probably follow the "save before action takes into place" pattern in this pane. We should probably run through this decision with Mia as well as include this button in the settings pane.

src/themer.tsx
```diff
@@ -6,39 +6,30 @@ import * as React from 'react';
 import { Fabric } from 'office-ui-fabric-react/lib/Fabric';
```
YingXue: seems like we would have some unit test debts to pay here later :)

src/app/settings/components/settingsPane.tsx
YingXue: I talked to Mia. She is ok putting the toggle in this page.
My suggestion is to make user click save before the changes apply. I looked at ADB, their language preference settings follows the same pattern. I think it also aligns with the accesibility guidelines a bit better.

src/app/devices/deviceContent/components/directMethod/__snapshots__/directMethod.spec.tsx.snap
montgomp: Especially since the required theme prop wasn't being passed. Thanks for catching things jest wasn't throwing on.

src/app/settings/components/settingsPane.tsx
montgomp: Have implemented as described above while moving to context-based theme changes.

montgomp:
Opening new PR as my branches in fork have gone sideways.

PR #130, opened 2019-10-16
Test gap filling

YingXue:
filled up some gaps in tests
removed snapshot on mounted component, only test snapshot on the component itself to remove a lot of unecessary details in the snapshot
test coverge is now 85.08%


src/app/devices/deviceContent/components/deviceContent.spec.tsx
```diff
@@ -0,0 +1,53 @@
...
+            ...overrides
+        };
+
+        return (
+            <DeviceContentComponent {...props} />
+        );
+    };
+
+    it('matches snapshot', () => {
+        testSnapshot(getComponent());
```
montgomp: would be good to add snapshot test after state changed for appMenuVisible

src/app/devices/deviceContent/components/deviceProperties/deviceProperties.spec.tsx
```diff
@@ -59,22 +61,36 @@ describe('components/devices/deviceProperties', () => {
...
-        expect(wrapper).toMatchSnapshot();
+    it('shows Shimmer while loading', () => {
+        const wrapper = mountWithLocalization(
+            getComponent(), false, true, [pathname]
+        );
         expect(wrapper.find(Shimmer)).toBeDefined();
     });
 
+    it('matches snapshot while interface cannot be found', () => {
+        testSnapshot(getComponent({isLoading: false, twinWithSchema: undefined}));
```
montgomp: When I run the tests locally, I'm seeing not covered branch for when it should be returning `<InterfaceNotFoundMessageBoxContainer/>` so I don't think this is doing overriding the props as you'd expect.

src/app/devices/deviceContent/components/deviceSettings/deviceSetttingsPerInterfacePerSetting.spec.tsx
```diff
@@ -54,7 +54,7 @@ describe('components/devices/deviceSettingsPerInterfacePerSetting', () => {
         settingModelDefinition: propertyModelDefinition,
         settingSchema: propertySchema};
 
-    it('matches snapshot when there is a writable property of simple type without sync status', () => {
+    it('renders when there is a writable property of simple type without sync status', () => {
```
montgomp: Will need more tests to cover displayName and description returning '--' and unit returning something other than '--'.

src/app/devices/deviceContent/components/shared/complexReportedFormPanel.spec.tsx
```diff
@@ -40,13 +40,10 @@ describe('complexReportedFormPanel', () => {
...
-        return mountWithLocalization(
-            <ComplexReportedFormPanel {...props} />
-        );
+        return (<ComplexReportedFormPanel {...props} />);
     };
 
     it('matches snapshot without twinWithSchema', () => {
-        const wrapper = getComponent();
-        expect(wrapper).toMatchSnapshot();
+        testSnapshot(getComponent());
```
montgomp: Would be good to add test where displayName is rendered as `'--'`

src/app/devices/deviceList/components/deviceList.spec.tsx
```diff
@@ -5,48 +5,81 @@
...
-        <DeviceListComponent
-            connectionString="connectionString"
-            devices={devices}
-            listDevices={jest.fn()}
-            deleteDevices={jest.fn()}
-            isFetching={false}
-        />);
-
-        expect(wrapper).toMatchSnapshot();
+        testSnapshot(getComponent());
```
montgomp: Would be good to test snapshots with `query.continuationTokens` and `query.currentPageIndex`

montgomp: Also with rendering `lastActivityTime` and `statusUpdatedTime` rendering as `'--'`

src/app/devices/deviceList/components/deviceListCell.spec.tsx
```diff
@@ -0,0 +1,48 @@
...
+    const getComponent = (overrides = {}) => {
+        const props = {
+            ...deviceListCellProps,
+            ...overrides,
+        };
+        return <DeviceListCell {...props} />;
+    };
+
+    it('matches snapshot while loading', () => {
+        testSnapshot(getComponent());
```
montgomp: Would be good to render additional snapshots with the various `context.t(ResourceKeys.deviceLists.noData)` branches being hit.

src/app/settings/components/repositoryLocationListItem.tsx
```diff
@@ -63,7 +64,7 @@ export default class RepositoryLocationListItem extends React.Component<Reposito
                         </div>
                         {item.repositoryLocationType !== REPOSITORY_LOCATION_TYPE.Public && <IconButton
                             className="remove-button"
-                            iconProps={{ iconName: 'cancel' }}
+                            iconProps={{ iconName: CANCEL }}
```
montgomp: I have a couple more tests on my machine we can add to the spec file for this component.

src/app/shared/utils/testHelpers.tsx
```diff
@@ -5,7 +5,9 @@
 import * as React from 'react';
 import { shallow, mount } from 'enzyme';
 import { MemoryRouter } from 'react-router-dom';
+import { Provider } from 'react-redux';
```
montgomp: we should ignore this file from test coverage and remove the `testWithLocalizationContextAndErrorBoundary` as it is not being called by any tests.

YingXue:
> We have a few more test scenarios to add to boost coverage but not something to block this PR on.

Definitely! Targeting 90% test coverage in the next wave :)

PR #131, opened 2019-10-17
fix buffer type event message

YingXue:


src/app/api/services/eventHubMessageHelper.ts
```diff
@@ -0,0 +1,26 @@
+/***********************************************************
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License
+ **********************************************************/
+import { Message } from '../models/messages';
+
+export const parseEventHubMessage = (message: Message): Message => {
+
+    if (!message) {
+        return null;
```
montgomp: This looks like it will return `null` if the message sent is `false`. Should be comparison to `undefined` or `null` or you can return `message` directly here.

montgomp: Never mind. I had forgotten that the value is stored in `.body`.

PR #132, opened 2019-10-17
Multiple hub cs

YingXue:
change textfield of hub connection string to a drop down, and allow users to save up to 5 strings.
refactored the settings pane and the login panel to use the same hub connection string componenet.
added tests.
image


src/app/constants/browserStorage.ts
```diff
@@ -3,7 +3,8 @@
  * Licensed under the MIT License
  **********************************************************/
 export const PRIVATE_REPO_CONNECTION_STRING_NAME = 'PRIVATE_REPO_CONNECTION_STRING_NAME';
-export const CONNECTION_STRING_NAME = 'CONNECTION_STRING_NAME';
+export const CONNECTION_STRING_LIST_MAX_LENGTH = 5;
+export const CONNECTION_STRING_NAME_LIST = 'CONNECTION_STRING_NAME'; // store hub connection strings in localStorage separated by comma
```
andrew-buckley: nit: Should this be CONNECTION_STRING_LIST_NAME?

YingXue: I wanted to change it as well, but I think we are "married" to old value now, as we can't break the experiecen for people who used the old version before

montgomp: I think we'd have _both_ `CONNECTION_STRING_NAME` to hold the selected connection string from the `CONNECTION_STRING_LIST` values.

src/app/css/_connectivityPane.scss
```diff
@@ -48,4 +48,14 @@
...
     }
-}
\ No newline at end of file
+}
+
+.connection-string-dropDown {
+    width: 95% !important;
+}
+
+.hidden {
```
montgomp: I think this is something that causes some HUGE accessibility nightmares of screens shifting focus.

YingXue: this is what we already do for masked text field: maskedCopyableTextField.tsx. I don't believe we got accesibility on portal. let me confirm with @c-ryan-k

src/app/constants/browserStorage.ts
YingXue: i don't think it is necessary though. because
1, it is duplicate information. the first in the list if what user currently selected, we only need the selector to extract information out
2. We would have to maintain both of the value in sync
So I am not going to add both of the value just for the sake of the naming convention, If necessary I can revert the constant's name back.

src/app/css/_connectivityPane.scss
c-ryan-k: Yeah, we experienced a similar issue with Firefox, instead of shifting things a bunch of pixels away to hide them, you can leave them in place but make them 'effectively invisible':
`height: 1px;`
`width: 1px;`
`opacity: 0;`

and for screen readers, etc - the elements that are "hidden" should be marked with `aria-hidden="true"`

src/app/constants/browserStorage.ts
YingXue: or if we are ok with breaking the users just once, we can rename it

montgomp: If we do, we should tag with Breaking Change.

YingXue: yeah that's true. probably not worth doing it?

src/app/css/_connectivityPane.scss
YingXue: updated to follow the best practice

src/app/constants/browserStorage.ts
montgomp: I'd still do it and just tag it. Better to do it early. And it's really a mild inconvenience of having to put the connection string back in. Not like it is taking away functionality.

PR #134, opened 2019-10-21
add a message bar indicating newer version to download if applies

YingXue:
image


src/localization/locales/en.json
```diff
@@ -292,6 +292,10 @@
                     }
                 }
             }
+        },
+        "messageBar": {
+            "message": "{{version}} has been release.",
```
c-ryan-k: nit: "released."

src/app/login/components/appVersionMessageBar.tsx
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    public componentDidMount() {
+        fetchLatestReleaseTagName().then(tagName => {
+            this.setState({latestReleaseVersion: tagName});
+        });
+    }
+
+    private readonly hasNewerRelease = () => {
+        if (!this.state || !this.state.latestReleaseVersion) { return false; }
```
montgomp: would `this.state` ever be falsey?

src/app/login/components/appVersionMessageBar.tsx
```diff
@@ -0,0 +1,57 @@
...
+        const latestVersion = this.state.latestReleaseVersion;
+        const versionTagPrefix = 'v';
+        const semanticVersion = latestVersion.startsWith(versionTagPrefix) ? latestVersion.split(versionTagPrefix)[1] : latestVersion;
```
montgomp: Looks like this could be replaced with `const semanticVersion = this.state.latestReleaseVersion.replace(/^v/);`, correct?

src/app/login/components/appVersionMessageBar.tsx
YingXue: this.state cannot be false.
But it did bring out a good point. this.state.latestReleaseVersion can possibly be in format which is not v + sematic.  Added a try catch block to not render the message bar in this case.

Issue #138, opened 2019-10-25
[BUG] Connecting to the public preview company repository


mattias-kindborg-axis-communications:
**Describe the bug**
I'm testing the public preview of IoT PnP, and have been reading [the documentation](https://docs.microsoft.com/en-us/azure/iot-pnp/). Azure IoT explorer is mentioned [here](https://docs.microsoft.com/en-us/azure/iot-pnp/quickstart-create-pnp-device#install-the-azure-iot-explorer), but when I try to connect to the repository I get an error.

**To Reproduce**
1. Start application
2. Enter the connection string, as described [here](https://docs.microsoft.com/en-us/azure/iot-pnp/quickstart-create-pnp-device#get-the-connection-string-for-your-company-model-repository)
3. Click on the 'Refresh' toolbar button

**Expected behavior**
I would expect to see the retrieved devices.

**Screenshots**
![image](https://user-images.githubusercontent.com/51745092/67484645-cd207980-f668-11e9-8344-06a9c1383d76.png)

**Desktop (please complete the following information):**
 - OS: Windows 10 Enterprise
 - OS Version: 1809
 - App version: 0.10.6

andrew-buckley:
Hi @mattias-kindborg-axis-communications. I'm wondering if you can check to see if there is any helpful information in the console/network logs `Help -> Toggle Developer Tools`.

mattias-kindborg-axis-communications:
![image](https://user-images.githubusercontent.com/51745092/67660402-03a70e80-f95f-11e9-9970-b531d23135b9.png)

YingXue:
From the screenshot above, I am guessing the connection string of the hub could be wrong, can you make sure if you copied the right connection string from portal?
You can copy it from: Azure Portal -> Your iot hub -> select 'Shared access policies' from the menu, and choose one policy that has the right permission:
![image](https://user-images.githubusercontent.com/5489222/67712735-cacc6100-f981-11e9-9a74-483bfcfe6ac6.png)

mattias-kindborg-axis-communications:
[The documentation](https://docs.microsoft.com/en-us/azure/iot-pnp/quickstart-create-pnp-device#get-the-connection-string-for-your-company-model-repository) tells me to copy the connection string from the public preview of the _Azure Certified for IoT portal_, not the IoT Hub.

![image](https://user-images.githubusercontent.com/51745092/67752234-f2c7cd00-fa32-11e9-8986-5860c5a8e355.png)

[...]

PR #139, opened 2019-10-28
fix error when it is not in json format

YingXue:
When response is not json, it throws error that looks like bug in the code: an example here: https://github.com/Azure/azure-iot-explorer/issues/138

Added a try catch with a bit of generic hep text
image

Some tests start to fail possibly due to the deprecation warnings start to become errors. Fixed in second commit.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -14,7 +14,7 @@ jobs:
         strategy:
             matrix:
                 node-version: [8.x, 10.x, 12.x]
-                os: [ubuntu-latest, windows-latest, macOS-latest]
+                os: [ubuntu-latest, windows-2016, macOS-latest]
 
         steps:
         - uses: actions/checkout@v1
@@ -23,6 +23,7 @@ jobs:
           with:
               node-version: ${{ matrix.node-version }}
         - name: npm install, build, and test
+          timeout-minutes: 10
           run: |
               npm install
               npm run build --if-present
--- a/src/app/api/services/devicesService.spec.ts
+++ b/src/app/api/services/devicesService.spec.ts
@@ -75,41 +75,15 @@ describe('deviceTwinService', () => {
             expect(DevicesService.fetchDeviceTwin(parameters)).toEqual(emptyPromise);
         });
 
-        it ('returns if connection string is not valid', () => {
-            expect(DevicesService.fetchDeviceTwin({...parameters, deviceId, connectionString: undefined})).toEqual(emptyPromise);
-            expect(DevicesService.fetchDeviceTwin({...parameters, deviceId, connectionString: 'SharedAccessKeyName=owner;SharedAccessKey=fakeKey='})).toEqual(emptyPromise);
+        it ('throws if connection string is not valid', async () => {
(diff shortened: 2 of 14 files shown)


andrew-buckley:
Is it always because the connection string is wrong?

YingXue:
> Is it always because the connection string is wrong? The error message seems to be more specific than generic.

Agree... I changed the wording to be slightly loose. Most of time, if the response from the API doesn't contain any json body (which is the case we are seeing error with no messages) are the case we will show this 'generic error', which mostly due to invalid connection string.

andrew-buckley:
> > Is it always because the connection string is wrong? The error message seems to be more specific than generic.
> 
> Agree... I changed the wording to be slightly loose. Most of time, if the response from the API doesn't contain any json body (which is the case we are seeing error with no messages) here, are the case we will show this 'generic error', which mostly due to invalid connection string.

Ideally we would have a regex check on the connection string and report specifically what is invalid about the connection string. Otherwise without a response from the server we don't *really* know what's going on.

YingXue:
> > > Is it always because the connection string is wrong? The error message seems to be more specific than generic.
> > 
> > 
> > Agree... I changed the wording to be slightly loose. Most of time, if the response from the API doesn't contain any json body (which is the case we are seeing error with no messages) here, are the case we will show this 'generic error', which mostly due to invalid connection string.
> 
> Ideally we would have a regex check on the connection string and report specifically what is invalid about the connection string. Otherwise without a response from the server we don't _really_ know what's going on.

I've updated the helper to check if it contains the repo id. But even with the regex check, we don't know if it is wrong or not. A example is, if a connection string is deleted from the portal, then user would not get any reponse back (which belongs to the cases I am handling here).

andrew-buckley:
> > > > Is it always because the connection string is wrong? The error message seems to be more specific than generic.
> > > 
> > > 
> > > Agree... I changed the wording to be slightly loose. Most of time, if the response from the API doesn't contain any json body (which is the case we are seeing error with no messages) here, are the case we will show this 'generic error', which mostly due to invalid connection string.
> > 
> > 
> > Ideally we would have a regex check on the connection string and report specifically what is invalid about the connection string. Otherwise without a response from the server we don't _really_ know what's going on.
> 
> I've updated the helper to check if it contains the repo id. But even with the regex check, we don't know if it is wrong or not. A example is, if a connection string is deleted from the portal, then user would not get any reponse back (which belongs to the cases I am handling here).
> 
> Also, those cases, they won't even fail at the api step, as it would fail the connection string text field validation.
> It is the cases that it cannot be pre-validated, and still the api has no response body, is where we want to show users of this error.

I guess what I'm saying is we only really know as much as what the API response is telling us. For the same reason you referenced above, we can't distinguish between the connection string being wrong, the connection string keys being removed, the hub being deleted, or any number of scenarios. So what I'm saying is it seems like we're trying to infer something specific from no information since the response body is empty which is think could be problematic. The information we do have is the connection string and the regex which we can easily validate. Which is why I prefer the new wording because if it turns out not to be a connection string problem we weren't determining with certainty that this was the issue.

PR #149, opened 2019-11-05
change list size and fix perf issue

YingXue:
Bump device list default size to 100
Before every single onChange on device id search box is calling setQuery on DeviceList component causing perf issue if device list size is increased. Now the logic has been changed to only setQuery when search icon is hit or user hit 'enter' key during device id search, so device list can be increased with no side effect.

A slight change is that, when user click 'Add query parameter' id search inbox would be disabled instead of cleared up, so the logic is easier, and user can also perserve the typing history when they removed all the queries.


andrew-buckley:
I wonder what the motivation is to switching page size to a new non-configurable value? Would it be better as a configurable option?

YingXue:
> I wonder what the motivation is to switching page size to a new non-configurable value? Would it be better as a configurable option?

This is sort of step 1 towards our future goal to make the size configurable. But the main content of this pr is to fix the perf issue of the query to enable us to make the size configurable in the future.

montgomp:
I'm seeing an issue where if I click "Add query parameter", it disables the "Query by device ID..." field and only re-enables if I remove (possibly many) clauses.

YingXue:
> I'm seeing an issue where if I click "Add query parameter", it disables the "Query by device ID..." field and only re-enables if I remove (possibly many) clauses.
> 
> This seems more cumbersome to use in this manner.

That is intended as in the description, so the user can still switch back to id search while it is perserving the history, instead of removing users' input (in a more predictive manner).

PR #151, opened 2019-11-07
Tooltip host

YingXue:
Couple key navigation related accessbility fixes:
1. Update office version. The previous version has a regression on dialog initial focus
2. Move mashead to layout.ts, as settings and notification was able to be discovered by tabbing
3. Use TooltipHost to wrap icon tooltips where tooltip needs to be shown while tabbing
4. Change callout icon's event from mouse event to onclick so it can be opened from keyboard 'space' key, and change focus back to icon when it is closed
5. A few text update

The above fixes address bug: 5660487 5659998 5652306 5660790 5660019 5660571

### DIFF
--- a/src/app/shared/components/maskedCopyableTextField.tsx
+++ b/src/app/shared/components/maskedCopyableTextField.tsx
@@ -6,6 +6,7 @@ import * as React from 'react';
 import { TranslationFunction } from 'i18next';
 import { IconButton } from 'office-ui-fabric-react/lib/Button';
 import { getId } from 'office-ui-fabric-react/lib/Utilities';
+import { TooltipHost } from 'office-ui-fabric-react/lib/Tooltip';
 import { ResourceKeys } from '../../../localization/resourceKeys';
 import LabelWithTooltip from './labelWithTooltip';
 import LabelWithRichCallout from './labelWithRichCallout';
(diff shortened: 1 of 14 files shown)


src/app/shared/components/maskedCopyableTextField.tsx
```diff
@@ -76,22 +77,24 @@ export class MaskedCopyableTextField extends React.Component<MaskedCopyableTextF
                         />
 
                         {allowMask &&
-                            <IconButton
-                                iconProps={hideContents ? { iconName: 'RedEye' } : { iconName: 'Hide' }}
-                                title={t(ResourceKeys.common.maskedCopyableTextField.toggleMask.label)}
-                                ariaLabel={t(ResourceKeys.common.maskedCopyableTextField.toggleMask.ariaLabel)}
-                                onClick={this.toggleDisplay}
-                            />
+                            <TooltipHost content={t(ResourceKeys.common.maskedCopyableTextField.toggleMask.label)}>
```
c-ryan-k: Quick note- is 
`ResourceKeys.common.maskedCopyableTextField.toggleMask.label` different than `ResourceKeys.common.maskedCopyableTextField.toggleMask.ariaLabel` ?

If you check out [the Fabric example](https://developer.microsoft.com/en-us/fabric#/controls/web/tooltip) it looks like we can avoid having duplicate labels for the controls with tooltips by using the `aria-describedby` attribute:

```
<TooltipHost content="This is the tooltip" id={this._hostId}>
   <DefaultButton aria-describedby={this._hostId}>Hover Over Me</DefaultButton>
</TooltipHost>
```

PR #153, opened 2019-11-08
fix query pill accessbility

YingXue:
Set focus on the first dropdown when new query is added
Set proper width on remove botton's focus
Fix border color (red when error, and all other time same as other border)

Bugs are addressed are: 5660526, 5660547


src/app/devices/deviceList/components/deviceQueryClause.tsx
```diff
@@ -65,7 +65,7 @@ export default class DeviceQueryClause extends React.PureComponent<DeviceQueryCl
     }
 
     private readonly shouldShowOperator = (parameterType: ParameterType) => {
-        return ParameterType.capabilityModelId !== parameterType && ParameterType.interfaceId !== parameterType;
+        return parameterType && ParameterType.capabilityModelId !== parameterType && ParameterType.interfaceId !== parameterType;
```
montgomp: Seems redundant. Was there a case where the existing comparisons didn't work as expected?

YingXue: The case when paremeterType is null is not covered. So three input field is renders by default, where showing two is what I have in mind as expected, because the first option (DCM) from the list renders two inputs so it feels like slightly less disruptive for the users.

PR #156, opened 2019-11-20
Disabled input

YingXue:
bug 5660862: Use disabled instead of readOnly to not allow tabbing
bug 5660862: annouce results when there is no devices. aslo added annouced to notifications.
bug 5660790: set focus back to 'copy' button after clicked
bump version


montgomp:
Requesting more info for 566082 (disabled vs readonly). Tab stop should exist for users to interact (copy).

PR #167, opened 2019-12-13
add module identity

YingXue:
1. Added add module ideneity functionality (which is very similar to add device logic)
2. Replaced reprecated componentWillReceiveProps with componentDidUpdate through the code base


src/app/devices/deviceContent/components/moduleIdentity/moduleIdentity.tsx
```diff
@@ -9,11 +9,12 @@ import { CommandBar } from 'office-ui-fabric-react/lib/CommandBar';
 import { LocalizationContextConsumer, LocalizationContextInterface } from '../../../../shared/contexts/localizationContext';
 import { ResourceKeys } from '../../../../../localization/resourceKeys';
 import { getDeviceIdFromQueryString } from '../../../../shared/utils/queryStringHelper';
-import { REFRESH } from '../../../../constants/iconNames';
+import { REFRESH, ArrayOperation } from '../../../../constants/iconNames';
```
montgomp: Constants should use SCREAMING_SNAKE_CASE

montgomp: Disregard. I just noticed this is an enum and not a single, static value.

YingXue: :) Sure. I've just merged master in.

PR #170, opened 2019-12-17
Update routes

chieftn:
Adopts url matching to decouple components from parent routes.


src/app/devices/deviceContent/components/deviceContentNav.spec.tsx
```diff
@@ -14,13 +14,16 @@ describe('components/devices/deviceContentNav', () => {
     const getComponent = (overrides = {}) => {
         const props = {
             deviceId: 'test',
+            history: jest.fn() as any, // tslint:disable-line:no-any
```
YingXue: I would recommend separate the three routerprops as another const for consistency with other tests.

PR #171, opened 2019-12-18
separate port for local development and electron app

YingXue:
1. Separate server port for local development and electron app, as we've always seeing 'port in use' error if we are trying to use the app in both ways. This PR is to get rid of that error.
2. Added an AppEnvironment constant, and used in telemetry page, so in hosted mode user fetch temetry once and it auto stops, while in electron mode, it doesn't auto stop.

_In the future, when we need to host the site, it should be as easy as adding another webpack and point the controller api endpoint to the hosted endpoint (hopefully)._

### DIFF
--- a/src/localization/locales/en.json
+++ b/src/localization/locales/en.json
@@ -442,7 +442,7 @@
             "showSystemProperties": "Show system properties",
             "start": "Start",
             "stop": "Stop",
-            "fetch": "Fetch"
+            "fetch": "Retrieve events"
         },
         "consumerGroups": {
             "label": "Consumer Group",
--- a/src/localization/locales/en.json
+++ b/src/localization/locales/en.json
@@ -441,7 +441,8 @@
             "clearEvents": "Clear events",
             "showSystemProperties": "Show system properties",
             "start": "Start",
-            "stop": "Stop"
+            "stop": "Stop",
+            "fetch": "Fetch"
         },
         "consumerGroups": {
             "label": "Consumer Group",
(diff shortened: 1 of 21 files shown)


src/server/serverElectron.ts
```diff
@@ -0,0 +1,35 @@
...
+    eventHubStopUri,
+    handleEventHubStopPostRequest,
+    modelRepoUri,
+    handleModelRepoPostRequest
+} from './serverHelper';
+
+const SERVER_PORT = 8081;
+const app = express();
+
+app.use(bodyParser.json());
```
montgomp: Please rename serverHelper.ts to serverBase.ts. Also, you can move all this logic into an init function that takes serverPort as an argument. Then serverLocal and serverElectron get refactored to:

```
import { init } from './serverBase';
const SERVER_PORT = 808x;
init(SERVER_PORT);
```

YingXue: makes total sense.

YingXue: Made a class and assign server_port in the constructor here :)

src/localization/locales/en.json
```diff
@@ -441,7 +441,8 @@
             "clearEvents": "Clear events",
             "showSystemProperties": "Show system properties",
             "start": "Start",
-            "stop": "Stop"
+            "stop": "Stop",
+            "fetch": "Fetch"
```
montgomp: "Fetch" seems a bit off-putting here. Would "Get events" or "Retrieve events" be better?

PR #173, opened 2019-12-18
Add logout and add tests

montgomp:
Feedback from security v-team to include a "log out" which is really just purging connection strings. Improve test coverage (themer).


src/themer.spec.tsx
```diff
@@ -0,0 +1,82 @@
...
+        );
+        expect(mount(component)).toMatchSnapshot();
+    });
+    it('matches snapshot with high contrast', () => {
+        localStorage.removeItem(THEME_SELECTION);
+        localStorage.setItem(HIGH_CONTRAST, 'true');
+        const component = (
+            <Themer />
+        );
+        expect(mount(component)).toMatchSnapshot();
```
YingXue: let's use testSnapshot() helper instead. the mounted component create unecessary details in the snapshot like font, color, classnames etc.

src/app/login/actions.ts
```diff
@@ -4,10 +4,12 @@
  **********************************************************/
 import actionCreatorFactory from 'typescript-fsa';
 import * as actionPrefixes from '../constants/actionPrefixes';
-import * as actionTypes from '../constants/actionTypes';
+import { LOG_OUT, SET_CONNECTION_STRING } from '../constants/actionTypes';
 
 const loginCreator = actionCreatorFactory(actionPrefixes.LOGIN);
-export const setConnectionStringAction = loginCreator<SetConnectionStringActionParameter>(actionTypes.SET_CONNECTION_STRING);
+export const setConnectionStringAction = loginCreator<SetConnectionStringActionParameter>(SET_CONNECTION_STRING);
+
```
YingXue: an unnecessary blank line here.

PR #174, opened 2019-12-18
Log out button and tests

montgomp:
Logout button per security v-team request.
Tests

*second attempt of PR (git reflog to the rescue)


src/themer.spec.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { THEME_SELECTION, HIGH_CONTRAST } from './app/constants/browserStorage';
+
+describe('themer', () => {
+    it('matches snapshot', () => {
+        localStorage.removeItem(THEME_SELECTION);
+        localStorage.removeItem(HIGH_CONTRAST);
+        const component = (
+            <Themer />
+        );
+        expect(mount(component)).toMatchSnapshot();
```
YingXue: let's use testSnapshot( ) test helper, which can help get rid of the unncessary details of the snapshot, like font, color and classnames.

montgomp: Moved to more expect calls based on state per offline conversation.

montgomp:
Abandoning PR. Looking a more sweeping retooling for the connection string management.

PR #178, opened 2019-12-23
Rkessler/sso store1

chieftn:
Adds Routing for Azure Resource
Indirection between connection string and active azure resource.


src/app/azureResource/actions.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+        };
+        expect(setActiveAzureResourceByConnectionStringAction(parameters)).toEqual({
+            payload: parameters,
+            type: 'AZURE_RESOURCES/SET_CONNECTION'
+        });
+    });
+});
+
+describe('setActiveAzureResourceByHostNameAction', () => {
+    it('returns AZURE_RESOURCES/SET_CONNECTION action object', () => {
```
YingXue: a small copy paste issue, should be SET_HOST here

src/app/azureResource/actions.ts
```diff
@@ -0,0 +1,27 @@
...
+import { SET } from '../constants/actionTypes';
+import { AzureResource } from './models/azureResource';
+
+export const AZURE_RESOURCES = 'AZURE_RESOURCES';
+export const BY_CONNECTION = '_CONNECTION';
+export const BY_HOSTNAME = '_HOST';
+
+const actionCreator = actionCreatorFactory(AZURE_RESOURCES);
+
+export interface SetActiveAzureResourceByConnectionStringActionParameters {
```
YingXue: is there any specific reason that we want to make 'hostName' a separate property? If we remove it, we can extract it out from connection string when we need it, so we don't need to pass it all the way down.

src/app/azureResource/components/azureResourceView.spec.tsx
```diff
@@ -0,0 +1,105 @@
...
+describe('AzureResourceView', () => {
+    it('matches snapshot when current resource is undefined', () => {
+        expect(shallow(
+            <AzureResourceView
+                activeAzureResource={undefined}
+                currentHostName="hostName"
+                currentUrl="url"
+                setActiveAzureResourceByHostName={jest.fn()}
+            />
+        )).toMatchSnapshot();
```
YingXue: is it possible to use the testSnapshot( ) helper here instead? We are trying to make the test pattern more consistent but haven't fully migrate all the tests which are still using pattern, but we are trying to make the new tests using the test helpers if possible.

src/app/azureResource/components/azureResourceView.tsx
```diff
@@ -0,0 +1,74 @@
...
+        if (activeAzureResource && activeAzureResource.hostName === currentHostName) {
+            return;
+        }
+
+        setActiveAzureResourceByHostName(currentHostName);
+    }, [currentHostName]); // tslint:disable-line:align
+
+    const renderContents = (): JSX.Element => {
+        if (!activeAzureResource) {
+            return (<></>);
```
YingXue: can NoMatchError component fit here?

src/app/azureResource/components/azureResourceViewContainer.spec.tsx
```diff
@@ -0,0 +1,30 @@
...
+import * as Redux from 'react-redux';
+import { shallow, mount } from 'enzyme';
+import { AzureResourceViewContainer, AzureResourceViewContainerProps } from './azureResourceViewContainer';
+
+describe('AzureResourceViewContainer', () => {
+    it('matches snapshot', () => {
+        jest.spyOn(Redux, 'useSelector').mockImplementation(() => 'active azure resource');
+        jest.spyOn(Redux, 'useDispatch').mockImplementation(jest.fn());
+
+        const routerprops: AzureResourceViewContainerProps = { // tslint:disable-line:no-any
```
YingXue: disable linter can be removed in this line

src/app/azureResource/state.ts
```diff
@@ -0,0 +1,15 @@
+/***********************************************************
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License
+ **********************************************************/
+import { AzureResource } from './models/azureResource';
+
+export interface AzureResourceStateInterface {
+    activeAzureResource: AzureResource | undefined;
```
YingXue: I am always puzzled by the decision between optional parameter vs the pattern above myself. Is there any difference that we would be mindful here?

src/app/connectionStrings/sagas/deleteConnectionStringSaga.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+
+        it('finishes', () => {
+            expect(addConnectionStringSagaGenerator.next()).toEqual({
+                done: true,
+            });
+        });
+    });
+
+    describe('performing no set action when no value retrieved', () => {
+        const addConnectionStringSagaGenerator = cloneableGenerator(deleteConnectionStringSaga)(deleteConnectionStringAction('connectionString1'));
```
YingXue: add -> delete

PR #195, opened 2020-01-10
Rkessler/add configuration settings (#10)

chieftn:
Expand configuration settings.


src/appConfig/appConfig.ts
```diff
@@ -0,0 +1,19 @@
...
+    Electron = 'electron'
+}
+
+export interface AppConfigInterface {
+    developmentMode: boolean;
+    hostMode: HostMode;
+    authMode: AuthMode;
+}
+
+export const appConfig = config as AppConfigInterface;
```
YingXue: I would prefer to add a helper here to extract the common logic like 'appConfig.hostMode === HostMode.Electron'  or 'appConfig.developmentMode' to provide some convinience. What do you say.

PR #198, opened 2020-01-13
Rkessler/remove extra wrapper properties source (#11)

chieftn:
Refactor wrapper types into single type.


src/app/api/models/synchronizationWrapper.ts
```diff
@@ -2,10 +2,8 @@
  * Copyright (c) Microsoft Corporation. All rights reserved.
  * Licensed under the MIT License
  **********************************************************/
-import { ModuleTwin } from './moduleTwin';
 import { SynchronizationStatus } from './synchronizationStatus';
-
-export interface ModuleIdentityTwinWrapper {
-    moduleIdentityTwin?: ModuleTwin;
+export interface SynchronizationWrapper<T> {
+    payload?: T;
```
YingXue: other than the name of property being 'payload' here, this is a great pattern. Maybe 'content' or something more general can be better. 'Payload' sounds a bit too API related.

PR #204, opened 2020-01-22
simplify breadcrumb by using recursion and route

YingXue:
Simplifies breadcrumb.
We still have some switch cases because we need to get device id and interface id from search parameters from the route, but everything else should now be generically handled without too many branching logic.


src/app/devices/deviceList/components/addDevice/components/addDevice.tsx
```diff
@@ -72,7 +71,7 @@ export default class AddDevice extends React.Component<AddDeviceActionProps & Ad
                 {(context: LocalizationContextInterface) => (
                      <form onSubmit={this.handleSave} className="view add-device">
                         <div className="view-header">
-                            <Route component={BreadcrumbContainer} />
+                            <Breadcrumb/>
```
asrudra: is not it a nested route inside deviceList? why do you need to mount breadcrumb here explicitly?

src/app/shared/components/__snapshots__/breadcrumb.spec.tsx.snap
```diff
@@ -1,168 +1,96 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`components/shared/breadcrumb matches snapshot #/ 1`] = `
-<StyledBreadcrumbBase
-  className="device-content-breadcrumb"
-  items={Array []}
-/>
+exports[`Breadcrumb snapshot matches snapshot 1`] = `
+<Fragment>
```
asrudra: this is not checking anything.. test with a particular route

src/app/shared/components/breadcrumb.tsx
```diff
@@ -3,165 +3,127 @@
...
-
-                if (index < pathComponents.length - 1) {
-                    const remainingPathComponents = pathComponents.slice(index + 1);
-                    items.push(...this.buildDigitalTwinDetails(remainingPathComponents, hubName, deviceId, interfaceId, t));
-                }
-                return;
+    // tslint:disable-next-line:cyclomatic-complexity
+    private renderBreadcrumbItem = (t: TranslationFunction, route: string | undefined) => {
+        if (route) {
+            switch (route) {
```
asrudra: so much better

src/app/shared/components/breadcrumb.tsx
```diff
@@ -3,165 +3,127 @@
...
+                case this.props.hostName:
+                    return this.renderTextItem(t(ResourceKeys.breadcrumb.hub, {hubName: this.getShortHubName(route)}));
+                case ROUTE_PARTS.DEVICE_DETAIL:
+                    return this.renderTextItem(getDeviceIdFromQueryString(this.props));
+                case ROUTE_PARTS.DIGITAL_TWINS:
+                    return this.renderTextItem(getInterfaceIdFromQueryString(this.props));
+                case ROUTE_PARTS.MODULE_IDENTITY:
+                    return this.renderLinkItem(this.getLocalizedKey(t, route), [ROUTE_PARAMS.DEVICE_ID]);
+                default:
+                    return this.renderLinkItem(this.getLocalizedKey(t, route));
```
asrudra: you are unnecessarily passing this 't' around all these functions.. this is a stateless component ..should be pretty easy to use localization hook.

src/app/shared/components/breadcrumb.tsx
```diff
@@ -3,165 +3,127 @@
...
-            basePath = `${basePath}/${paths[0]}`;
-            entries.push({
-                href: paths.length !== 1 ? `${basePath}?${query}` : '',
-                key: paths[0],
-                text: t((ResourceKeys.deviceContent.navBar as any)[paths[0]]) // tslint:disable-line:no-any
+    private renderLinkItem = (linkContent: string, searchParams?: ROUTE_PARAMS[]) => {
+        const { match } = this.props;
+        let href = match.url;
+        if (searchParams) {
+            href += '/';
```
asrudra: you should remove this otherwise default case in the switch will also have an extra '/' at the end.

src/app/shared/components/breadcrumb.tsx
```diff
@@ -3,165 +3,127 @@
...
+            searchParams.forEach(para => {
```
asrudra: i don't understand this . should not we have the switch on Route content ? how would you handle this if any future route needs more than one route params? (Ex: a/b/c?foo=bar&bar=baz)

src/app/shared/components/breadcrumb.tsx
YingXue: default case has no trailing slash, only when search parameters are present

src/app/shared/components/breadcrumb.tsx
YingXue: this is exactly why this function takes in an array. This is helper function arleady based on route, and it passed the seach params into this helper to make the logic reusable.

src/app/shared/components/__snapshots__/breadcrumb.spec.tsx.snap
YingXue: the snapshot is not doing anything, but it is checking the switch cases for example which is adding up to the coverage percentage..

PR #206, opened 2020-01-23
Rkessler/add auth dependency (#12)

chieftn:
Adds SSO infrastructure.


src/app/api/models/authorizationRuleNotFound.Error.ts
```diff
@@ -0,0 +1,14 @@
+/***********************************************************
```
YingXue: An interesting naming of the file. Is it intended since we have an error folder already?

src/app/iotHub/models/sharedAccessSignatureAuthorizationRule.ts
```diff
@@ -2,10 +2,11 @@
...
  * Licensed under the MIT License
  **********************************************************/
+import { AccessRights } from './accessRights';
 
 export interface SharedAccessSignatureAuthorizationRule {
-    accessRights: string;
     keyName: string;
-    primaryKey: string;
-    secondaryKey: string;
+    primaryKey?: string;
```
YingXue: can keys be optional?

src/app/iotHub/sagas/getSharedAccessSignatureAuthorizationRulesSaga.ts
```diff
@@ -0,0 +1,58 @@
...
+import { getSharedAccessSignatureAuthorizationRules } from '../services/iotHubService';
+import { AzureResourceIdentifier } from '../../azureResourceIdentifier/models/azureResourceIdentifier';
+import { SharedAccessSignatureAuthorizationRule } from '../models/sharedAccessSignatureAuthorizationRule';
+import { executeAzureResourceManagementTokenRequest } from '../../login/services/authService';
+import { appConfig } from '../../../appConfig/appConfig';
+import { StateInterface } from '../../shared/redux/state';
+import { CacheWrapper } from '../../api/models/cacheWrapper';
+
+const cacheInMinutes = 4;
+const secondsPerMinute = 60;
```
YingXue: 60 and 1000 are two constants in utils.ts which can be exported and used here.

src/app/azureResourceIdentifier/services/azureResourceIdentifierService.ts
```diff
@@ -0,0 +1,110 @@
...
+        method: HTTP_OPERATION_TYPES.Post
+    };
+    const response = await fetch(resourceUrl, serviceRequestParams);
+    if (!response.ok) {
+        throw new HttpError(response.status);
+    }
+
+    const responseBody = await response.json() as AzureResourceIdentifierQueryResult;
+    const azureResourceIdentifiers = (responseBody.data && responseBody.data.rows) ?
+        responseBody.data.rows.map(row => mapPropertyArrayToObject(azureResourceManagementQueryFields, row) as AzureResourceIdentifier) :
```
YingXue: a fun journey to undersand this logic personally. Is it that we query all the fields, and the returned array 'row' here would have everything in order, and we then map it back to an oject with keys being the qeury fields? Wondering if the helper which checks the length boundary is just an extra check or we actually run into the situation that the two lengths would be different.

src/app/api/models/authorizationRuleNotFound.Error.ts
chieftn: yeah -- this looks like a mistake.  will fix.

src/app/iotHub/models/sharedAccessSignatureAuthorizationRule.ts
chieftn: pulled this from hub -- believe they are optional.

src/app/iotHub/sagas/getSharedAccessSignatureAuthorizationRulesSaga.ts
chieftn: np -- going to move these to constants as exported values (currently not exported);

src/app/azureResourceIdentifier/services/azureResourceIdentifierService.ts
chieftn: the helper takes the desiredFields as the canonical value -- if the number of entries in the row exceeds the number of desired fields -- the value is not written to the returned data object.  Conversely, the desiredField is returned with 'undefined' if the value array is out of bounds.

PR #209, opened 2020-02-01
Fix port 404 bug

YingXue:
Throw specific error when the port explorer needed is in use, and allow users to leverage environment varible to solve this issue for now.
Electron would read env and put the value into local storage for client to use.

### DIFF
--- a/src/app/devices/deviceList/sagas/listDeviceSaga.ts
+++ b/src/app/devices/deviceList/sagas/listDeviceSaga.ts
@@ -40,7 +40,7 @@ export function* listDevicesSaga(action: Action<DeviceQuery>) {
         else {
             // tslint:disable-next-line:cyclomatic-complexity
             text = {
-                translationKey: (action.payload && action.payload.clauses) ? ResourceKeys.notifications.getDeviceListQueryGenericErrorHelp : ResourceKeys.notifications.getDeviceListGenericErrorHelp
+                translationKey: ResourceKeys.notifications.getDeviceListGenericErrorHelp
             };
         }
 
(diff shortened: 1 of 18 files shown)


src/app/devices/deviceList/sagas/listDeviceSaga.ts
```diff
@@ -21,16 +24,26 @@ export function* listDevicesSaga(action: Action<DeviceQuery>) {
...
     } catch (error) {
-        const text = error && error.message ? {
-            translationKey: ResourceKeys.notifications.getDeviceListOnError,
-            translationOptions: {
-              error: error.message,
-          },
-        } :
-        {
+        let text;
+        if (error.name === ERROR_TYPES.PORT_IS_IN_USE) {
```
digimaun: We discussed restructuring here.

src/server/serverElectron.ts
```diff
@@ -5,4 +5,11 @@
 import ServerBase from './serverBase';
 
 const SERVER_PORT = 8081;
-(new ServerBase(SERVER_PORT)).init();
+// try to read from environment variable to check if the users have set up a specific port to use
+try {
+    const customPort = parseInt(process.env.AZURE_IOT_EXPLORER_PORT); // tslint:disable-line:radix
+    (new ServerBase(customPort || SERVER_PORT)).init();
+}
+catch {
```
digimaun: Might be more clear to catch the specific error(s)/exception that would constitute another attempt at deploying the server rather than any exception.

YingXue: yeah true. We would need to revisit this approach (thinking when a related issue got raised maybe) for future improvement. Kinda want to get rid of manual step as a whole.

PR #221, opened 2020-02-11
Component name

YingXue:
This PR's majority is renaming: rename interface name to component name

Also, since we allow multiple components to implement the same interface, using interface in breadcrumb and url in breadcrum and url is not unique enough. Change it to use component name instead.

The selector also got some cleanup and update, since component is already in url.

image


src/app/devices/deviceContent/components/digitalTwin/digitalTwinInterfaces.tsx
```diff
@@ -7,7 +7,7 @@ import { useSelector, useDispatch } from 'react-redux';
 import { Label } from 'office-ui-fabric-react/lib/Label';
 import { CommandBar, ICommandBarItemProps } from 'office-ui-fabric-react/lib/CommandBar';
 import { RouteComponentProps, NavLink } from 'react-router-dom';
-import { getDigitalTwinInterfaceIdToNameMapSelector, getDigitalTwinDcmNameSelector, getDigitalTwinInterfacePropertiesWrapperSelector } from '../../selectors';
+import { getDigitalTwincomponentNameAndIdsSelector, getDigitalTwinDcmNameSelector, getDigitalTwinInterfacePropertiesWrapperSelector } from '../../selectors';
```
montgomp: Component should be capitalized

src/app/devices/deviceContent/components/digitalTwin/digitalTwinInterfaces.tsx
```diff
@@ -22,7 +22,7 @@ import '../../../../css/_devicePnPDetailList.scss';
 
 export interface DigitalTwinInterfacesProps extends RouteComponentProps{
     isLoading: boolean;
-    idToNameMap: Map<string, string>;
+    nameToIds: object;
```
montgomp: What is the data structure here? Is there anything better than object that we can us?

src/app/devices/deviceContent/components/digitalTwin/digitalTwinInterfaces.tsx
```diff
@@ -36,18 +36,22 @@ export const DigitalTwinInterfaces: React.FC<DigitalTwinInterfacesProps> = props
...
 
-    const navLinks: JSX.Element[] = [];
-    if (props.idToNameMap && props.idToNameMap.size > 0) {
-        props.idToNameMap.forEach((componentName, interfaceName) => {
-            const link = `${url}${ROUTE_PARTS.DIGITAL_TWINS_DETAIL}/${ROUTE_PARTS.INTERFACES}/?${ROUTE_PARAMS.DEVICE_ID}=${encodeURIComponent(deviceId)}&${ROUTE_PARAMS.INTERFACE_ID}=${interfaceName}`;
-            navLinks.push(
+    let navLinks: JSX.Element[] = [];
+    if (props.nameToIds) {
+        navLinks = Object.keys(props.nameToIds).map(componentName => {
+            const interfaceId = (props.nameToIds as any)[componentName]; // tslint:disable-line:no-any
```
montgomp: This usage makes me believe that nameToIds is still a `Map<string, string>`

src/app/devices/deviceContent/selectors.ts
```diff
@@ -58,7 +58,7 @@ export const getDigitalTwinInterfacePropertiesSelector = (state: StateInterface)
         state.deviceContentState.digitalTwinInterfaceProperties.payload;
 };
 
-export const getDigitalTwinInterfaceNameAndIdsSelector = createSelector(
+export const getDigitalTwincomponentNameAndIdsSelector = createSelector(
```
montgomp: Component should have capital C

src/app/devices/deviceContent/components/digitalTwin/digitalTwinInterfaces.tsx
YingXue: nameToIds is the data sctructure returned by service, so I am accessing it as the object. Map value be retrieved by nameToIds.get('key'). The 'any' here gets rid of the index linter warning here.
It simply looks like `{'name1`: 'id1', 'name2': 'id2'}`

YingXue: It totally reminded me that this page has no unit test. Going to add one. Also just talked to Eddie, he suggested putting the interfaces and components into a table. I will do so in the update.

src/app/devices/deviceContent/components/digitalTwin/digitalTwinInterfaces.tsx
YingXue: Please see the comments below.

PR #229, opened 2020-02-21
rename digital twin to pnp

YingXue:
As titled, rename string (reviewd by ChrisG). The files names are left as is. Will get to them if needed in the future.
The nav is also flattened, as PnP become simply one of the nav item.
image


README.md
```diff
@@ -1,5 +1,5 @@
 
-# Azure IoT Plug and Play (PnP) Device Explorer
+# Azure IoT explorer
```
montgomp: Thank you for this. I'm not sure how we've overlooked it for so long.

src/app/devices/deviceContent/components/deviceContentNav.tsx
```diff
@@ -46,41 +45,34 @@ export default class DeviceContentNavComponent extends React.Component<DeviceCon
...
                 key: NAV_LINK_ITEM_PNP,
                 name: context.t((ResourceKeys.deviceContent.navBar as any)[NAV_LINK_ITEM_PNP]), // tslint:disable-line:no-any
                 url: `#${url}/${NAV_LINK_ITEM_PNP}/?${ROUTE_PARAMS.DEVICE_ID}=${encodeURIComponent(deviceId)}`
             }];
-            groups.push({
-                links: pnpNavLinks,
-                name: context.t(ResourceKeys.deviceContent.navBar.pnp),
-            });
+
+            navLinks = [...navLinks, ...pnpNavLinks];
```
montgomp: since pnpNavLinks is a single item, line 48 can be `const navLinks  =...` and this line can be `navLinks.push(pnpNavLinks);`

YingXue:
> I think the renaming of the Files/Components from Digital Twin to Plug and Play should definitely come sooner rather than later. It will have much confusion with route parts and strings saying Plug and Play and the file/component names saying Digital Twin. Can we add a backlog item for making that change as a distinct PR?

Sure. Created a task to keep track of the file renames

Issue #228, opened 2020-02-20
Add support for viewing telemetry routed to a custom endpoint


damiandiaz76:
Currently, the "Telemetry" tab views events sent to the built in endpoint "events".  I configured message routing to route telemetry events to a different event hub.  It would be nice to be able to see my routed telemetry events in explorer.

YingXue:
Hi @damiandiaz76 
The new release is out with this feature support, please give it a try and let us know if it doens't work for you: https://github.com/Azure/azure-iot-explorer/releases/tag/v0.10.16

damiandiaz76:
I tried the custom event hub feature.  It started to "Receiving... events" but when events came in, the right side window closed. 
<img width="1195" alt="Screen Shot 2020-04-15 at 4 45 46 PM" src="https://user-images.githubusercontent.com/19311187/79396203-9b61de80-7f38-11ea-9bf4-2592e4c8752b.png">

damiandiaz76:
Looks like starting with build 0.10.15, telemetry from the built in endpoint also behaves the same way: when events come in, the telemetry closes.

PR #237, opened 2020-03-11
Eventhub

YingXue:
- Addressing request #228: support users monitoring events from custom event hub
- Fix package vunerablity

### DIFF
--- a/src/app/api/services/devicesService.ts
+++ b/src/app/api/services/devicesService.ts
@@ -343,21 +343,18 @@ export const deleteDevices = async (parameters: DeleteDevicesParameters) => {
 // tslint:disable-next-line:cyclomatic-complexity
 export const monitorEvents = async (parameters: MonitorEventsParameters): Promise<Message[]> => {
     try {
-        if (!parameters.hubConnectionString) {
+        if (!parameters.hubConnectionString && !parameters.customEventHubConnectionString) {
             return;
         }
 
         const requestParameters = {
-            connectionString: parameters.hubConnectionString,
-            consumerGroup: parameters.consumerGroup,
-            deviceId: parameters.deviceId,
-            fetchSystemProperties: parameters.fetchSystemProperties,
+            ...parameters,
             startTime: parameters.startTime && parameters.startTime.toISOString()
         };
 
         const response = await request(EVENTHUB_MONITOR_ENDPOINT, requestParameters);
         const messages = await response.json() as Message[];
-        return  messages && messages.length !== 0 && messages.map(message => parseEventHubMessage(message)) || [];
+        return  messages && messages.length && messages.length !== 0 && messages.map(message => parseEventHubMessage(message)) || [];
     } catch (error) {
         throw error;
     }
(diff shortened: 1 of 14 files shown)


src/app/api/services/devicesService.ts
```diff
@@ -343,21 +343,18 @@ export const deleteDevices = async (parameters: DeleteDevicesParameters) => {
...
-            deviceId: parameters.deviceId,
-            fetchSystemProperties: parameters.fetchSystemProperties,
+            ...parameters,
             startTime: parameters.startTime && parameters.startTime.toISOString()
         };
 
         const response = await request(EVENTHUB_MONITOR_ENDPOINT, requestParameters);
         const messages = await response.json() as Message[];
-        return  messages && messages.length !== 0 && messages.map(message => parseEventHubMessage(message)) || [];
+        return  messages && messages.length && messages.length !== 0 && messages.map(message => parseEventHubMessage(message)) || [];
```
c-ryan-k: nit: not sure the extra check for ```.length !== 0``` is necessary - if messages is null, the first check will fail, and if it's length is 0 then won't ```messages.length``` be falsy?

src/app/css/_deviceEvents.scss
```diff
@@ -37,10 +37,33 @@
...
+        }
+        width: 80%;
     }
 
     .toggle-button {
         padding-left: 24px;
     }
+
+    .scrollable-telemetry {
+        height: calc(100vh - 400px);
```
c-ryan-k: Suggestion - for items with calculated dimensions that use a hard-coded number value of pixels (also see _devicePnpDetailList.scss), we might want to store these in variables to be used elsewhere, or at least so we know why they're offset by 400/360px

src/server/serverBase.ts
```diff
@@ -209,10 +210,20 @@ export const addPropertiesToCloudToDeviceMessage = (message: CloudToDeviceMessag
...
+                body.customEventHubConnectionString && body.customEventHubConnectionString !== connectionString)
+            {
+
+                client = body.customEventHubConnectionString ?
+                    await EventHubClient.createFromConnectionString(body.customEventHubConnectionString) :
+                    await EventHubClient.createFromIotHubConnectionString(body.hubConnectionString);
+
+                connectionString = body.customEventHubConnectionString ?
+                    body.customEventHubConnectionString :
+                    body.customEventHubConnectionString;
```
c-ryan-k: should this be ```body.hubConnectionString``` ?

src/app/api/services/devicesService.ts
YingXue: the messages can be an array or an object. Withought the message.length check, it could possibly throw in .map, as message.length indeed !== 0 because it is undefined.

src/app/css/_deviceEvents.scss
YingXue: I agree. I think we really need a better to handle these all together. We have a lot of inline scrolls, I was wondering if we can a more automatically way to handle these.

PR #240, opened 2020-03-23
Minimist vulnerability fix

c-ryan-k:
Temporarily add a force-resolution task to prevent extract-zip from installing older version of mkdirp and therefore minimist < 1.2.3


c-ryan-k:
FYI - I added the github actions update due to encountering this a few times after one of our checks timed out and I tried to re-run:
https://github.com/actions/checkout/issues/23

PR #241, opened 2020-03-25
fix css

YingXue:
With the previous update of packages, ms grid system stopped working and broke the pnp grid styles.
Moved it to use flex.

### DIFF
--- a/src/app/css/_devicePnpDetailList.scss
+++ b/src/app/css/_devicePnpDetailList.scss
@@ -23,6 +23,7 @@
 }
 
 .pnp-detail-list {
+    margin-left: 15px;
     @include themify($themes) {
         background-color: themed('backgroundColor');
     }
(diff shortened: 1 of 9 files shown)


src/app/css/_devicePnpDetailList.scss
```diff
@@ -50,6 +51,26 @@
     }
 }
 
+.flex-grid-row {
+    display: flex;
+    justify-content: space-between;
+}
+.flex-grid-row .col-sm4 {
```
c-ryan-k: Can we simplify this as:
```
.flex-grid-row {
    ...
    .col-sm4 {
    }
    .col-sm3 {
    }
    ...
}
```
?

YingXue: yeah sure

PR #242, opened 2020-03-25
Rkessler/flex test (#13)

chieftn:
Provide reusable application frame for masthead, breadcrumb et.al.
Streamline css rules

### DIFF
--- a/src/app/css/_applicationFrame.scss
+++ b/src/app/css/_applicationFrame.scss
@@ -33,7 +33,6 @@
     }
 
     .navigation {
-        min-height: 90px;
         width: 100%;
         grid-row: 2;
         grid-column: 1;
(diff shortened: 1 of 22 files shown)


src/app/css/_applicationFrame.scss
```diff
@@ -7,53 +7,45 @@
...
+        width: 100%;
         grid-row: 1;
         grid-column: 1;
         -ms-grid-row: 1;
         -ms-grid-column: 1;
     }
-    .content {
+
+    .navigation {
+        min-height: 90px;
```
YingXue: Not sure if I missed, wonder if you would like to leave it min-height or make it fixed height

chieftn: yes -- looks like an oversight -- will fix.

PR #250, opened 2020-04-07
use @id to match dtdl not names

YingXue:
When retrieving model definition from local folder, instead of asking users to name local files following strict rules, we will parse json files in the folder, and try to match @id

### DIFF
--- a/src/localization/locales/en.json
+++ b/src/localization/locales/en.json
@@ -79,7 +79,7 @@
                 "local": {
                     "labelInElectron":"Local folder",
                     "labelInBrowser":"Local folder (This is only enabled in the desktop version from: https://github.com/Azure/azure-iot-explorer/releases)",
-                    "infoText": "Use your local folder as a model repository. The interface files you wish to use should have a json extension and a file name equal to the Interface ID with all colons (':') removed. For example: urn:contoso:com:EnvironmentalSensor:1 should be named as urncontosocomEnvironmentalSensor1.json",
+                    "infoText": "Use your local folder as a model repository. The model definition files you wish to use should have a json extension. We will retrieve model definition by dtmi through searching the matching '@id' from the files, and the first matching file content would be retrieved.",
                     "textBoxLabel": "Local folder path",
                     "placeholder": "f:/"
                 }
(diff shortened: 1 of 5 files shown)


src/localization/locales/en.json
```diff
@@ -79,7 +79,7 @@
                 "local": {
                     "labelInElectron":"Local folder",
                     "labelInBrowser":"Local folder (This is only enabled in the desktop version from: https://github.com/Azure/azure-iot-explorer/releases)",
-                    "infoText": "Use your local folder as a model repository. The interface files you wish to use should have a json extension and a file name equal to the Interface ID with all colons (':') removed. For example: urn:contoso:com:EnvironmentalSensor:1 should be named as urncontosocomEnvironmentalSensor1.json",
+                    "infoText": "Use your local folder as a model repository. The model definition files you wish to use should have a json extension. We will retrieve model definition by dtmi through searching the matching '@id' from the files, and the first matching file content would be retrieved.",
```
andrew-buckley, line 82: Consider avoiding acronyms here. Also I'm a little confused by the wording. Does this mean that it will return the first interface definition where the IDs match?

src/server/serverBase.ts
```diff
@@ -51,33 +52,59 @@ export default class ServerBase {
...
+    for (const fileName of fileNames) {
+        if (isFileExtensionJson(fileName)) {
+            try {
+                const data = fs.readFileSync(`${filePath}/${fileName}`, 'utf-8');
+                if (JSON.parse(data)['@id'].toString() === expectedFileName) {
+                    return data;
+                }
+            }
+            catch {
+                // swallow error
```
andrew-buckley, line 96: This could be swallowing a file read error or a JSON parsing error.

src/localization/locales/en.json
YingXue: Yes. Added the full name for dtmi, and @id is what is defined in dtdl 
image
I am kinda struggling to find the right wording for this.

andrew-buckley: Yeah I like that! "The first JSON file with an `@id` property matching the digital twin model identifier (DTMI) will be retrieved."

PR #252, opened 2020-04-08
Rkessler/resource view (#14)

chieftn:
Adds resource view.


src/app/azureResources/components/azureResourcesContainer.tsx
```diff
@@ -0,0 +1,7 @@
+import * as React from 'react';
```
YingXue, line 1: Learning from Ashish's comment for new FC patterns, we would adopt the pattern that new separate container files is needed but simply put it inside of the FC view.

src/app/connectionStrings/components/connectionStringsViewContainer.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { useSelector, useDispatch } from 'react-redux';
+import { RouteComponentProps } from 'react-router-dom';
+import { StateInterface } from '../../shared/redux/state';
+import { upsertConnectionStringAction, deleteConnectionStringAction, setConnectionStringsAction } from '../actions';
+import { setActiveAzureResourceByConnectionStringAction } from '../../azureResource/actions';
+import { ROUTE_PARTS } from '../../constants/routes';
+import { ConnectionStringsView } from './connectionStringsView';
+import { formatConnectionStrings } from '../../shared/utils/hubConnectionStringHelper';
+
+export const ConnectionStringsViewContainer: React.FC<RouteComponentProps> = props => {
```
YingXue, line 15: Same comment here to combine the files. If dispatch effects if not tested, I would move useDispatch to the fc component since they don't need to be passed from the container. I was also asked to do the same for useSelector, but I haven't make that change yet due to not able to test it.
So combine the files is what I would suggest for now, and we can take another chance to refactor all useDispatch and useSelector together.

chieftn: will do.  I'm actually not a fan of using the useDispatch / useSelector directly in the component.  I feel this tightly couples the component while also complicating tests.

YingXue: i had the same feeling to be honest. I am ok with going forward with only combining the files at this point. For useDispatch and useSelector, especially useDispatch i always had the issue with having shadowed parameter name. But consistency does matter for code base. Wondering if we should have a wider discussion as a team.

PR #254, opened 2020-04-09
Folder picker

YingXue:
A friendly folder picker. Instead of asking users to manually enter folder path, give users a folder picker.
image

image
As any browser side file picker would not render absolute path due to security concerns, it is achieved by having controls pull subfolders when user intiate a click including start picking, click on a folder name, or click back button to navigate to parent folder

Also revert a enum value back from last PR. As the enum value is stored in local storage, changing it would result user data lose.


src/localization/locales/en.json
```diff
@@ -80,8 +80,21 @@
...
+                        "command": {
+                            "openPicker": "Pick a folder",
+                            "navigateToParent": "Navigate back to parent folder",
+                            "select": "Select",
+                            "cancel": "Cancel"
+                        },
+                        "dialog": {
+                            "title": "Select a folder as a local model repository",
+                            "subText": "Selected folder: {{folder}}",
+                            "noFolderFoundText": "No folders found.",
```
asrudra, line 94: No folder found

src/app/shared/utils/utils.ts
```diff
@@ -41,3 +41,31 @@ export const validateModuleIdentityName = (key: string): boolean => {
...
+    return navigator.platform.indexOf('Win') > -1;
+};
+
+export const getRootFolder = (): string => {
+    return isPlatformWindows() ? null : '/';
+};
+
+// tslint:disable-next-line:cyclomatic-complexity
+export const getParentFolder = (currentFolder: string): string => {
+    const i = currentFolder.lastIndexOf('/');
```
asrudra, line 55: please consider renaming this

src/app/settings/components/repositoryLocationListItem.tsx
```diff
@@ -51,22 +76,116 @@ export default class RepositoryLocationListItem extends React.Component<Reposito
...
+                        text={context.t(ResourceKeys.settings.modelDefinitions.repositoryTypes.local.folderPicker.command.openPicker)}
+                        ariaLabel={context.t(ResourceKeys.settings.modelDefinitions.repositoryTypes.local.folderPicker.command.openPicker)}
+                        onClick={this.onShowFolderPicker}
+                    />
+                    {this.renderFolderPicker(context)}
+                </>
+            );
+    }
+
+    private readonly onShowFolderPicker = () => {
```
asrudra, line 112: why are these functions readonly ? these fns are changing state

YingXue: There use to have a linter in this repo that would yell if the functions are not readonly. Didn't pay attention when did this rule get removed, so I was blindly c&p. Let me update it.

src/localization/locales/en.json
YingXue: plural seems more common. If you try to open folder from vs code, the file explorer would show 'no items found'.

PR #256, opened 2020-04-15
Patch dt

YingXue:
- Implement patch dt api, and remove any residuals of dt interface property concept
- Explictly make required property to be null in json schema, so our ui won't flag error when user is trying to clear the value of writable properties


src/app/devices/deviceContent/sagas/digitalTwinSaga.ts
```diff
@@ -16,10 +20,63 @@ export function* getDigitalTwinSaga(action: Action<string>) {
...
+                    },
+                },
+                type: NotificationType.success
+            }));
+            yield put(patchDigitalTwinAction.done({params: action.payload, result: digitalTwin}));
+        }
+        else if (responseCode === DataPlaneStatusCode.SuccessLowerBound) {
+            yield put(addNotificationAction.started({
+                text: {
+                    translationKey: ResourceKeys.notifications.patchDigitalTwinOnSuccess,
```
montgomp, line 57: Only differs from the above by the value of translationKey, should be able to combine these two blocks of code.

YingXue: Yeah totally.

PR #259, opened 2020-04-21
Rkessler/semantic units

chieftn:
Adding semantic units support.


package.json
```diff
@@ -41,6 +41,7 @@
     "clean": "IF EXIST .\\dist RMDIR /Q /S .\\dist",
     "clean:linux": "rm --recursive -f ./dist",
     "localization": "tsc ./scripts/composeLocalizationKeys.ts --skipLibCheck && node ./scripts/composeLocalizationKeys.js",
+    "import-semantic-units": "tsc ./scripts/importSemanticUnitTypes.ts --skipLibCheck && node ./scripts/importSemanticUnitTypes.js",
```
YingXue, line 44: Let me confirm if I've understand it correctly. This 'import-semantic-units' is not included in any npm build script, because the output json is already included in the source code.
This script is for the future in case we need re-generate the json.

chieftn: exactly.  I suspect the they'll do updates over time -- like adding localization.

PR #260, opened 2020-04-22
Update json schema adaptor; integrate service update for json patch

YingXue:
- remove tooltip from json schema form. This has now become duplicate information and hence not necessary to show anymore
- fix map in object data converter. When there was no value in digital twin, couple edge case was not properly checked
- update dt service and selectors to deal with new patch update

### DIFF
--- a/src/app/api/models/interfaceJsonParserOutput.ts
+++ b/src/app/api/models/interfaceJsonParserOutput.ts
@@ -19,8 +19,9 @@ export interface ParsedJsonSchema {
 }
 
 export interface ParsedCommandSchema {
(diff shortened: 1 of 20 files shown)


src/app/api/models/interfaceJsonParserOutput.ts
```diff
@@ -19,8 +19,9 @@ export interface ParsedJsonSchema {
 }
 
 export interface ParsedCommandSchema {
-    description: string;
     name: string;
+
```
chieftn, line 23: extra space?

PR #266, opened 2020-04-27
Reusable schema

YingXue:
Implement dtdl reusable schema.
The **jsonSchemaAdaptor.ts** has switched from a set of utility funciton to an instantiable class. The reason being, during class instantiation, the class can store and parse the 'schemas' section of dtdl for future uses (example in **mockModelDefinition.ts**).
Most of the implemetation stays the same, while **parseInterfaceContentSchemaHelper** is extracted out from the old parseInterfacePropertyHelper, because of reusable schema definition in dtdl. The main change is in line 129, which utilize the 'reuse' capablity of json schema https://json-schema.org/understanding-json-schema/structuring.html#id1

### DIFF
--- a/src/app/api/models/modelDefinition.ts
+++ b/src/app/api/models/modelDefinition.ts
@@ -10,6 +10,7 @@ export interface ModelDefinition {
     contents: Array<PropertyContent | CommandContent | TelemetryContent | ComponentContent>;
     description?: string | object;
     displayName?: string | object;
+    schemas?: Array<ObjectSchema | MapSchema | EnumSchema>;
 }
 
 export interface PropertyContent extends ContentBase {
@@ -38,7 +39,6 @@ interface ContentBase {
     comment?: string | object;
     description?: string | object;
     displayName?: string | object;
-    displayUnit?: string;
     unit?: string;
 }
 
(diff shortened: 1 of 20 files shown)


src/app/api/models/modelDefinition.ts
```diff
@@ -38,7 +39,6 @@ interface ContentBase {
     comment?: string | object;
     description?: string | object;
     displayName?: string | object;
-    displayUnit?: string;
```
chieftn, line 41: Whoops.  Nice catch.

src/app/api/models/modelDefinition.ts
```diff
@@ -51,18 +51,22 @@ export interface Schema {
 
 export interface EnumSchema {
     '@type': string;
-    enumValues: Array<{ displayName: string | object, name: string, enumValue: number}>;
+    valueSchema: string;
+    enumValues: Array<{ displayName: string | object, name: string, enumValue: number | string}>;
```
chieftn, line 55: wondering if this should be a first class interface rather than anonymous one.

src/app/shared/utils/twinAndJsonSchemaDataConverter.ts
```diff
@@ -107,7 +107,7 @@ export const findPathsTowardsMapType = (
 };
 
 export const getNumberOfMapsInSchema = (settingSchema: ParsedJsonSchema): number => {
-    const hasMatch = JSON.stringify(settingSchema).match(/additionalProperties/g);
+    const hasMatch = settingSchema && JSON.stringify(settingSchema).match(/additionalProperties/g);
```
chieftn, line 110: notice a number of these checks have been removed.  Can you help me understand why they are no longer required?

YingXue: Can you point me to where these has been removed? I am adding the null check here, not trying to remove anything.

PR #287, opened 2020-05-13
render telemetry row with empty body

YingXue:
1. As titled. if the telemetry body is empty, still render a row in pnp telemetry page.
image

2. If there is any invalid json when we resolve from local folder, info users what are the files:
image


src/server/serverBase.ts
```diff
@@ -103,12 +103,12 @@ const findMatchingFile = (filePath: string, fileNames: string[], expectedFileNam
...
             catch {
-                errorsCaught ++; // swallow error and continue the loop
+                filesWithParsingError.push(fileName); // swallow error and continue the loop
             }
         }
     }
-    if (errorsCaught > 0) {
-        throw new Error();
+    if (filesWithParsingError.length > 0) {
+        throw new Error(filesWithParsingError.join(','));
```
c-ryan-k, line 111: Really minor nit - since this error is surfaced to the user, I think we should add a space to the join string so it appears as `a.json, b.json` instead of `a.json,b.json`

PR #295, opened 2020-06-03
Devicetwin refactor

YingXue:


src/app/devices/addDevice/state.ts
```diff
@@ -1,3 +1,7 @@
+/***********************************************************
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License
+ **********************************************************/
```
asrudra, line 4: do we have to add this on every file?

src/app/devices/deviceContent/deviceTwin/sagas.ts/getDeviceTwinSaga.spec.ts
```diff
@@ -0,0 +1,86 @@
+/***********************************************************
```
asrudra, line 1: folder name sagas.ts ?

src/app/devices/addDevice/state.ts
YingXue: According to the policy, yes. Maybe something like this is worth looking into: https://stackoverflow.com/questions/38823319/automatically-add-copyright-banner-in-vs-code

PR #297, opened 2020-06-05
Device identity

YingXue:
Refactor identity:
Device content create the local state and pass props to nav and device identity


asrudra:
rebase with master please

PR #303, opened 2020-06-08
Remove redux and reselect. Note: azureResource folder and iotHub folder are removed.  When need to bring them back, this is the commit to refer to

YingXue:


asrudra:
remove react-redux

PR #310, opened 2020-06-11
Updated react-i18next to remove create-react-context

c-ryan-k:


package.json
```diff
@@ -90,7 +90,7 @@
     "react": "16.8.6",
     "react-collapsible": "2.3.2",
     "react-dom": "16.8.6",
-    "react-i18next": "8.1.0",
+    "react-i18next": "^11.5.0",
```
YingXue, line 93: let's use exact version for this one

c-ryan-k: Oh, right, thanks - didn't notice that, will fix.

PR #313, opened 2020-06-17
fix bug when switch hubs and others

YingXue:
- Update:
  - PnP component landing page to show text when device is not pnp

- Bug fixes:
  - Make setConnectionStringsAction async, because before we already redirected to the list view, so everything got unmounted before the code to update connection strings in local storage get called.
  - Fix toggle button on device nav always set to be true
  - Fix breadcrumb where hostname retrival is async
  - Fix missing 500 case for local model resolution (for example a folder was previous set but got deleted)
  - Warning removal (deprecated ariaLabel on Link)
- Style:
  - Updated home view nav and content alignment to be consisten with device detail


src/app/shared/components/breadcrumb.tsx
```diff
@@ -14,12 +14,12 @@ import '../../css/_breadcrumb.scss';
...
 export const Breadcrumb: React.FC = () => {
     const [ hostName, setHostName ] = React.useState<string>('');
+    const connectionString = localStorage.getItem(ACTIVE_CONNECTION_STRING);
 
     React.useEffect(() => {
-        const connectionString = localStorage.getItem(ACTIVE_CONNECTION_STRING);
         const host = getConnectionInfoFromConnectionString(connectionString).hostName;
         setHostName(host);
-    },              []);
+    },              [connectionString]);
```
asrudra, line 22: nice 👍

PR #319, opened 2020-06-30
debug console for useReactAsyncReducer

asrudra:
image

### DIFF
--- a/src/app/connectionStrings/components/connectionStringsView.tsx
+++ b/src/app/connectionStrings/components/connectionStringsView.tsx
@@ -30,7 +30,7 @@ export const ConnectionStringsView: React.FC = () => {
     const { t } = useTranslation();
     const history = useHistory();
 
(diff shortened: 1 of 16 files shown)


src/app/connectionStrings/components/connectionStringsView.tsx
```diff
@@ -30,7 +30,7 @@ export const ConnectionStringsView: React.FC = () => {
     const { t } = useTranslation();
     const history = useHistory();
 
-    const [ localState, dispatch ] = useAsyncSagaReducer(connectionStringsReducer, connectionStringsSaga, connectionStringsStateInitial());
+    const [ localState, dispatch ] = useAsyncSagaReducer(connectionStringsReducer, connectionStringsSaga, connectionStringsStateInitial(), 'connectionStringsState');
```
YingXue, line 33: would it be a good idea to make a const in state.ts and use it here?

PR #329, opened 2020-07-14
bug bash fixes round 2

YingXue:
1. event useState refactor. the setState with ... is causing some intermittent issues on the state. moved all local state to seperate variable.
2. validation refactor to deal with map type in two different cases. Skip all or only skip certain error message with 'array' keyword.
3. checkbox false value fix

### DIFF
--- a/src/app/devices/deviceEvents/components/deviceEvents.spec.tsx
+++ b/src/app/devices/deviceEvents/components/deviceEvents.spec.tsx
@@ -84,22 +84,11 @@ describe('components/devices/deviceEvents', () => {
             }
         }];
 
-        const initialState = {
-            consumerGroup: DEFAULT_CONSUMER_GROUP,
-            customEventHubConnectionString: undefined,
-            customEventHubName: undefined,
-            events,
-            hasMore: false,
-            loading: false,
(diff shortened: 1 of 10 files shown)


src/app/devices/deviceEvents/components/deviceEvents.spec.tsx
```diff
@@ -84,22 +84,11 @@ describe('components/devices/deviceEvents', () => {
...
-            monitoringData: false,
-            showSystemProperties: false,
-            startTime: new Date(new Date().getTime() - MILLISECONDS_IN_MINUTE), // set start time to one minute ago
-            synchronizationStatus: SynchronizationStatus.initialized,
-            useBuiltInEventHub: true
-        };
         const realUseState = React.useState;
-        jest.spyOn(React, 'useState').mockImplementationOnce(() => realUseState(initialState));
+        jest.spyOn(React, 'useState').mockImplementationOnce(() => realUseState(DEFAULT_CONSUMER_GROUP));
+        jest.spyOn(React, 'useState').mockImplementationOnce(() => realUseState(undefined));
```
asrudra, line 89: can we 'mockImplementation' multiple times in one line?

PR #334, opened 2020-07-20
Merge telemetry

YingXue:
- Merge telemetry page and pnp telelemetry page as they two have a lot in common
  - Moved 'events' to local state and removed unnecessary complexity introduced by infinite scroll (package also removed)
  - System property is no longer retrieved different from api level, but become a client side toggle
  - Paid some styling debt, to use different height with different elements present on the page, to user as much as possible for events monitoring
- Fixed map patch data, where patch doesn't like null value
- Added extra regex check
- Using pivot to replace custom built control


src/app/devices/deviceEvents/reducers.ts
```diff
@@ -0,0 +1,58 @@
...
+            // filter overlaped messages returned from event hub
+            filteredMessages = messages.filter(message => message.enqueuedTime > state.payload[0].enqueuedTime);
+        }
+        return state.merge({
+            payload: [...filteredMessages, ...state.payload],
+            synchronizationStatus: SynchronizationStatus.fetched
+        });
+    })
+    .case(startEventsMonitoringAction.failed, (state: DeviceEventsStateType) => {
+        return state.merge({
```
asrudra, line 35: setIn would be crisp

PR #337, opened 2020-07-22
String update per request; get rid of ruler of json editor; friendly error message for nodejs conflicting port

YingXue:


public/electron.ts
```diff
@@ -157,6 +157,14 @@ else {
     app.on('ready', () => {
         createWindow();
         createMenu();
+        process.on('uncaughtException', err => {
+            const messageBoxOptions = {
+                    message: err && err.message || 'Something went wrong. Please try restart the app.',
```
c-ryan-k, line 162:
```suggestion
                    message: err && err.message || 'Something went wrong. Please try restarting the app.',
```

src/server/serverBase.ts
```diff
@@ -56,7 +56,11 @@ export class ServerBase {
         app.get(readFileUri, handleReadFileRequest);
         app.get(getDirectoriesUri, handleGetDirectoriesRequest);
 
-        app.listen(this.port);
+        app.listen(this.port).on('error', () => { throw new Error(
+           `Failed to start the app on port ${this.port} as it is in use.
+            You can still view static pages, but request cannot be made to the services if the port is still occupied.
```
c-ryan-k, line 61:
```suggestion
            You can still view static pages, but requests cannot be made to the services if the port is still occupied.
```

PR #356, opened 2020-09-04
retrieve inline component as flattened model definition

YingXue:
When component is defined inline, we are showing a composite interface id on the ui, `{interfaceid}/{inlineSchemaId}`
And thus, when we are retrieving corresponding model from local and repo, we retrieve interface with matching` {interfaceId} `and then fetch the corresponding component definition of `{inlineSchemaId}` as a flattened model definition

### DIFF
--- a/src/app/api/models/modelDefinition.ts
+++ b/src/app/api/models/modelDefinition.ts
@@ -5,6 +5,12 @@
 export type ComplexSchema = ObjectSchema | MapSchema | EnumSchema | ArraySchema;
 export type ContentTypeUnion = PropertyContent | CommandContent | TelemetryContent | ComponentContent;
 
+interface ModelSchema {
+    '@type': string;
+    '@id': string;
+    contents: ContentTypeUnion[];
+}
+
 export interface ModelDefinition {
     '@context': string;
     '@id': string;
@@ -14,11 +20,7 @@ export interface ModelDefinition {
     description?: string | object;
     displayName?: string | object;
     schemas?: ComplexSchema[];
-    schema?: {
-        '@type': string;
-        '@id': string;
-        contents: ContentTypeUnion[];
-    };
+    schema?: ModelSchema;
 }
 
 export interface PropertyContent extends ContentBase {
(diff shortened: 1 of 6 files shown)


montgomp (commented):
I think "splited" should be "split" throughout getModelDefinitionSaga.ts

src/app/api/models/modelDefinition.ts
```diff
@@ -31,7 +37,11 @@ export interface TelemetryContent extends ContentBase{
 }
 
 export interface ComponentContent extends ContentBase{
-    schema: string;
+    schema: string | {
+        '@type': string;
+        '@id': string;
+        contents: ContentTypeUnion[];
+    };
```
asrudra, line 44: should create a separate interface for this object to avoid repetition.

src/app/devices/pnp/sagas/getModelDefinitionSaga.ts
```diff
@@ -75,24 +75,45 @@ export function* validateModelDefinitionHelper(modelDefinition: ModelDefinition,
     }
 }
 
-export function* getModelDefinitionFromPublicRepo(action: Action<GetModelDefinitionActionParameters>, location: RepositoryLocationSettings) {
+export function* getSplitedInterfaceId(fullName: string) {
+    // when component definition is inline, interfaceId is compose of parent file name and inline schema id concatenated with a slash
+    return fullName.split('/');
+}
```
asrudra, line 81: why is this a generator function?

src/app/devices/pnp/sagas/getModelDefinitionSaga.ts
```diff
@@ -75,24 +75,45 @@ export function* validateModelDefinitionHelper(modelDefinition: ModelDefinition,
...
+    if (splitedInterfaceId.length === 1) {
+        return model;
+    }
+    else {
+        // for inline component, the flattened model is defined under contents array's with matching schema @id
+        const components = model.contents.filter((content: any) => // tslint:disable-line: no-any
+            content['@type'] === 'Component' && typeof content.schema !== 'string' && content.schema['@id'] === splitedInterfaceId[1]);
+        return components[0];
+    }
+}
```
asrudra, line 93: this also can be a normal fn

asrudra (changes requested):

asrudra (approved):

PR #361, opened 2020-10-01
Connect to new model repo

YingXue:
Allow users to configure to public model repo, as well as allow users to configure their own repo url
image


src/app/constants/apiConstants.ts
```diff
@@ -22,7 +22,7 @@ export const MODEL_ID_REF = '/ref:modelId?';
 export const MODEL_ID = 'modelId=';
 export const API_VERSION = 'api-version=';
 export const AND = '&';
-export const PUBLIC_REPO_HOSTNAME = 'repo.azureiotrepository.com';
+export const PUBLIC_REPO_HOSTNAME = 'devicemodeltest.azureedge.net';
```
digimaun, line 25: Point in time?

digimaun (dismissed):

src/app/constants/apiConstants.ts
montgomp: I wouldn't use the test url if this is going to be released. Would recommend `devicemodels.azure.com`

YingXue: This is what I saw in the wiki of model repo. Let me change it. I am also thinking have this pr only approved but not merged, if we need to get anything out before 10/16.

digimaun (dismissed):

PR #377, opened 2020-12-11
Security

YingXue:
Per security requests, we are adding an expiry (one year from now) for any hub connection string added.
There are several scenarios covered:
1. For existing users, on home page, they would see an warning notification notifying that their previously stored connection strings have been removed
2. For new strings added, it would be good for one year without user input change
3. For new strings close to expiration (threshold of 3 days), an warning message would be displayed like below
image
4. If the user ignores or didn't see this error message, the specific connection strings would have been removed the next time user log on, and the user would also get a warning notification


src/localization/locales/en.json
```diff
@@ -214,7 +214,8 @@
                 "label": "Connection String",
                 "ariaLabel": "Connection String"
             }
-        }
+        },
+        "expirationWarning": "Your hub connection string is expiring in less than {{numberOfDays}} day(s). Please edit the connection string (by copying and pasting again) to extend, or you would have to re-add it once it expires."
```
iluican, line 218: "would have" -> "will have"

src/app/shared/utils/hubConnectionStringHelper.ts
```diff
@@ -40,3 +41,23 @@ export const isValidEventHubConnectionString = (connectionString: string): boole
     const pattern = new RegExp('^Endpoint=sb://.*;SharedAccessKeyName=.*;SharedAccessKey=.*$');
     return pattern.test(connectionString);
 };
+
+export const getDaysBeforeHubConnectionStringExpires = (connectionStringWithExpiry: ConnectionStringWithExpiry) => {
+    // tslint:disable-next-line: no-magic-numbers
+    const millisecondsPerDay = 24 * 60 * 60 * 1000;
```
montgomp, line 47: Would be cleaner to do a single magic number here to avoid calculating every execution.

montgomp (changes requested):

src/app/constants/browserStorage.ts
```diff
@@ -3,7 +3,10 @@
  * Licensed under the MIT License
  **********************************************************/
 export const CONNECTION_STRING_LIST_MAX_LENGTH = 10;
-export const CONNECTION_STRING_NAME_LIST = 'CONNECTION_STRING_NAME_LIST'; // store hub connection strings in localStorage separated by comma
+export const CONNECTION_STRING_EXPIRATION_IN_YEAR = 1;
+export const CONNECTION_STRING_EXPIRATION_WARNING_IN_DAYS = 3;
+export const CONNECTION_STRING_NAME_LIST = 'CONNECTION_STRING_NAME_LIST'; // DEPRECATED: store hub connection strings in localStorage separated by comma
```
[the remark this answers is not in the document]
YingXue: That is also something I am not able to decide, so I went with a new list instead. The code might be dead once we have some telemetry showing there is no more active users using old versions, but for now, I don't have a good answer regarding to the timeline.

src/app/shared/utils/hubConnectionStringHelper.ts
YingXue: sure

src/app/constants/browserStorage.ts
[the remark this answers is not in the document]
YingXue: As dicussed with Paul of line, we will keep using the old key and add a try parsing logic, as it is more future proven (if we want to change the structure again, we need this parsing logic anyways) and we don't want to keep switching to the new keys.

src/app/connectionStrings/sagas/getConnectionStringsSaga.ts
```diff
@@ -3,15 +3,50 @@
...
+        {
+            // connection strings stored can not be parsed to ConnectionStringWithExpiry Array
+            // Currently it is possible that user has old format connection strings stored
+            yield call(raiseNotificationToast, {
+                text: {
+                    translationKey: ResourceKeys.notifications.connectionStringsWithoutExpiryRemovalWarning
+                },
+                type: NotificationType.warning
+            });
+            localStorage.setItem(CONNECTION_STRING_NAME_LIST, '');
```
montgomp, line 48: yes, exactly. Something bad happened. Let's get to where we can start new.

montgomp (approved):

PR #402, opened 2021-03-03
fall back to controller when debug round 2

YingXue:
make telemetry call and public model repo call fall back to controller during debug

### DIFF
--- a/src/app/api/legacy/localRepoServiceHandlers.ts
+++ b/src/app/api/legacy/localRepoServiceHandlers.ts
@@ -8,7 +8,7 @@ import { READ_FILE, CONTROLLER_API_ENDPOINT, DataPlaneStatusCode, GET_DIRECTORIE
 import { ModelDefinitionNotFound } from '../models/modelDefinitionNotFoundError';
 import { ModelDefinitionNotValidJsonError } from '../models/modelDefinitionNotValidJsonError';
 
-export class LocalRepoServiceHandler implements DirectoryInterface, ModelRepositoryInterface {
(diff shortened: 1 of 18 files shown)


src/app/api/legacy/localRepoServiceHandlers.ts
```diff
@@ -8,7 +8,7 @@ import { READ_FILE, CONTROLLER_API_ENDPOINT, DataPlaneStatusCode, GET_DIRECTORIE
 import { ModelDefinitionNotFound } from '../models/modelDefinitionNotFoundError';
 import { ModelDefinitionNotValidJsonError } from '../models/modelDefinitionNotValidJsonError';
 
-export class LocalRepoServiceHandler implements DirectoryInterface, ModelRepositoryInterface {
+export class LocalRepoServiceHandlers implements DirectoryInterface, ModelRepositoryInterface {
```
chieftn, line 11: The plural here seems a little weird -- may give wrong impression.  We can always split the handlers into two types if you are concerned about the (s) with respect to DirectoryInterface / ModelRepositoryInterface

src/app/api/shared/interfaceUtils.spec.ts
```diff
@@ -50,9 +50,9 @@ describe('browserSettingsApi', () => {
 });
 
 describe('getDeviceInterface', () => {
-    it('throws exception when mode is not electron', () => {
+    it('does not throw exception when mode is not electron', () => {
```
chieftn, line 53: I think this should probably be reframed as an assertion like 'call factory x' or removed outright.

chieftn (approved):

Issue #203, opened 2020-01-21
[Feature request]: Unable to see any module twin properties in module twin with latest release for iot edge module twins


initmahesh:
Using the latest release cannot see any module twin for ot edge modules.
Is there a plan to add support for edge module twin in Azure Iot Explorer.

so that we can update Iot Edge module twin from the tool.

**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

YingXue:
Hi,
The latest release only contains module identity which is a non-edge device related feature. There are plans to support edge but it's a bit further down the road.

PR #416, opened 2021-03-11
Module2

YingXue:
Add command and telemetry to module pnp, and raw telemetry page to module
Unhide module identities page for edge device #203

### DIFF
--- a/src/server/serverBase.ts
+++ b/src/server/serverBase.ts
@@ -2,6 +2,7 @@
  * Copyright (c) Microsoft Corporation. All rights reserved.
  * Licensed under the MIT License
  **********************************************************/
+// this file is the legacy controller for local development, until we move server side code to use electron's IPC pattern and enable electron hot reloading
 import * as fs from 'fs';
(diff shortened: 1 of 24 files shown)


src/server/serverBase.ts
```diff
@@ -2,6 +2,7 @@
  * Copyright (c) Microsoft Corporation. All rights reserved.
  * Licensed under the MIT License
  **********************************************************/
+// this file is the legacy controller for local development, until we move server side code to use electron's IPC pattern and enable electron hot reloading
```
YingXue, line 5: as the comment mentioned, this file is legacy, so there are some duplicate code if compare this file to EventHubHandler.ts
We haven't found a good way to enable electron local debugging during the RBAC refactor, so it is currently using serverBase in local development mode, while in electron mode using EventHubHandler
A bit more info in wiki: https://github.com/Azure/azure-iot-explorer/wiki/Dev-manual#how-to-debug-electron-main-dev

asrudra (approved):

PR #463, opened 2021-08-05
Ying/update event hub sdk

YingXue:
Update event hub sdk version to get rid of security vulnerablitiy
While doing so, we discovered that the signuare of the SDK introduced quite a few breaking changes.

- We are now converting our hub cs to be a event hub compatible string to create the consumer client.
- Eventhub event pulling is now a subscribe model. There is one draw back which is, we can only clear the message array (result array) after the monitor call from the UI has been returned. So when user click 'clear messages', it's possible that the previous 'getEvents' call is still filling up the result array, while we already cleared the events in our redux state, so we feel like the ui is 'one step behind' after the first button click.
- We can easily get all events perserved within retention without their new EventPosition parameter.

Side Note: since we are using IPC in electron and the legacy node in hosted (developer experience) you will see identitcal code in two places, serverBase and the eventHandler.

### General Guidelines

### DIFF
--- a/src/server/eventHubHelper.ts
+++ b/src/server/eventHubHelper.ts
@@ -1,0 +1,83 @@
+
+/***********************************************************
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License
+ **********************************************************/
+import { AmqpError, isAmqpError, Connection, ReceiverEvents, parseConnectionString } from "rhea-promise";
+import * as crypto from "crypto";
+
+// The following helper functions are directly copied from:
+// https://raw.githubusercontent.com/Azure/azure-sdk-for-js/main/sdk/eventhub/event-hubs/samples/v5/typescript/src/iothubConnectionString.ts
+export const convertIotHubToEventHubsConnectionString =  async (connectionString: string): Promise<string> =>
+{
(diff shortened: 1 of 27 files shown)


src/server/eventHubHelper.ts
```diff
@@ -0,0 +1,83 @@
...
+    // Generate a token to authenticate to the service.
+    // The code for generateSasToken can be found at https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-security#security-tokens
+    const token = generateSasToken(`${HostName}/messages/events`, SharedAccessKey, SharedAccessKeyName, 5);
+  
+    const connection = new Connection({
+        transport: "tls",
+        host: HostName,
+        hostname: HostName,
+        username: `${SharedAccessKeyName}@sas.root.${iotHubName}`,
+        port: 5671,
```
montgomp, line 40: Do you think port number should be in a constants file?

montgomp (approved):

src/server/eventHubHelper.ts
YingXue: ah i didn't even notice this, was a pure copy paste. Not sure if it would cause any issue if there is a port conflict, guess we can leave it and figure out?

Issue #477, opened 2021-10-29
[BUG] Unable to stream Telemetry Events


themegaphoenix:
**Describe the bug**
Events are not being streamed and shown in the Telemetry tab. Works fine on 0.14.5

**To Reproduce**
1. Clone https://github.com/Azure/azure-iot-sdk-python/tree/master/azure-iot-device/samples/pnp
2. Start it up and send telemetry events
3. Go to the Devices and then Telemetry and start streaming 

**Expected behavior**
Events show up

**Desktop (please complete the following information):**
 - OS: Windows
 - OS Version 10 
 - App version 0.14.6

**Additional context**
Works fine on 0.14.5

YingXue:
Hi @themegaphoenix 
Thank you very much for reporting the issue! We will release a new version to address the issue. Meanwhile, please fall back to the old version. Sorry for the inconvinience.

YingXue:
Fixed in release 0.14.7

smanni-et:
> Fixed in release 0.14.7

@YingXue the bug is still there in 0.14.7 as reported also in #481 

Rollback to 0.14.5 or previous is the only way to have the telemetry stream working.

Issue #475, opened 2021-10-22
[BUG] Wrong Fonts in ubuntu


rido-min:
Running in Ubuntu the fonts dont seem right:

![image](https://user-images.githubusercontent.com/14916339/138395467-98f395b7-a366-466a-8723-23ca19f21ae0.png)

YingXue:
from this site, there are several fonts recommended. We are also updating to Arial and see if that gets the app a better look: http://rigaux.org/font-family-compatibility-between-linux-.html

PR #478, opened 2021-10-29
fix events bug

YingXue:
Fixing #477 where we were comparing strings to date object
Addressing #475 to use a font that is similiar enough on various OS

### DIFF
--- a/src/app/css/_application.scss
+++ b/src/app/css/_application.scss
@@ -38,7 +38,7 @@
     }
 }
 .app {
-    font-family: "Arial" !important;
+    font-family: "Segoe UI", "Arial" !important;
     font-size: 14px !important;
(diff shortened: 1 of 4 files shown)


src/app/css/_application.scss
```diff
@@ -38,7 +38,7 @@
     }
 }
 .app {
-    font-family: "Segoe UI" !important;
+    font-family: "Arial" !important;
```
montgomp, line 41:
```suggestion
    font-family: "Segoe UI", "Arial" !important;
```
To allow using Segoe UI if it is installed and Arial as a fallback

montgomp (approved):
Let me know what you think of the font family suggestion.

montgomp (approved):

PR #513, opened 2022-07-12
Telemetry for page views - count and visit time

shahm5:
Added Azure App Insights logging to keep track of page views and page view times
---
This project has adopted the [Microsoft Open Source Code of Conduct](https://opensource.microsoft.com/codeofconduct/). For more information see the [Code of Conduct FAQ](https://opensource.microsoft.com/codeofconduct/faq/) or contact [opencode@microsoft.com](mailto:opencode@microsoft.com) with any additional questions or comments.

Thank you for contributing to the Azure IoT Explorer!

This checklist is used to make sure that common guidelines for a pull request are followed.

### General Guidelines


src/app/constants/telemetry.ts
```diff
@@ -0,0 +1,31 @@
+/***********************************************************
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License
+ **********************************************************/
+export enum TELEMETRY_PAGE_NAMES {
+    HUBS = 'hub_connections',
+    CLOUD_TO_DEVICE_MESSAGE = 'cloud_to_device_message',
+    DEVICE_TELEMETRY = 'device_telemetry', // same for device vs pnp telemetry vs module telemetry
```
YingXue, line 8: I appreciate the fact that you are trying to group similar funcitonal pages together, but we would like to separate the pnp and non pnp events page, so we have an idea of the actual usage data.

package.json
```diff
@@ -71,6 +71,7 @@
   "dependencies": {
     "@azure/event-hubs": "5.6.0",
     "@fluentui/react": "8.20.2",
+    "@microsoft/applicationinsights-web": "^2.8.4",
```
YingXue, line 74: that's use the eactly version of the package to be consist with other packages.

YingXue (changes requested):

YingXue:
> This does not seem to be right fix for test, as now we have duplicates in app.config and app.*.config. We have two things to do here:
> 
> 1. I assume that we have verified manually that using app.*.config before this commit does make things functional, the app is choosing the right string respectively
> 2. Look into how Jest can mock up the config values to get rid of the errors.

No worries on the dups of the reply. All you would need to do is, simply write appConfig.telemetryConnString = 'someteststring' in all the tests where it's saying 'missing instrument key'. That's my understanding, please check if that works, and revert the last change.

YingXue (commented):
LGTM. Let's update the version in package.json and ship it!

src/app/shared/appTelemetry/appInsightsClient.ts
```diff
@@ -0,0 +1,24 @@
...
+                autoTrackPageVisitTime: true,
+                connectionString: appConfig.telemetryConnString
+            } });
+            try {
+                appInsights.loadAppInsights();
+                appInsights.trackEvent({name: `INIT`}, {type: 'init'});
+            } catch (e) {
+                // tslint:disable-next-line:no-console
+                console.log(e);
+                AppInsightsClient.instance = appInsights;
```
YingXue, line 18: this value assignment looks to be in the wrong place. Are we only assigning instance in catch clause?

PR #516, opened 2022-07-25
revert event hub sdk update

YingXue:
Because so many people has reported telemetry not working consistently, and we cannot repro on our side, I am reverting the sdk update to previous version which is well tested and stable, and this was the pr :https://github.com/Azure/azure-iot-explorer/pull/463

### General Guidelines


src/app/devices/deviceEvents/components/deviceEvents.tsx
```diff
@@ -71,47 +70,51 @@ export const DeviceEvents: React.FC = () => {
...
     // simulation specific
     const [showSimulationPanel, setShowSimulationPanel] = React.useState(false);
 
-    React.useEffect(() => {
-        if (componentName) {
-            AppInsightsClient.getInstance()?.trackPageView({name: TELEMETRY_PAGE_NAMES.PNP_TELEMETRY});
-        } else {
-            AppInsightsClient.getInstance()?.trackPageView({name: TELEMETRY_PAGE_NAMES.DEVICE_TELEMETRY});
-        }
-    }, []); // tslint:disable-line: align
```
c-ryan-k, line 88: Is it ok to leave this in?
It looks like we still have `componentName` populated so I think we should be good to leave the telemetry in?

YingXue: Thanks for catching it! I copied the old file from the older commits, and forgot what was newly added. If we remove it, there won't be any views on this page. Phew!

c-ryan-k (approved):

PR #603, opened 2023-01-24
update per sdl review; fix color contrast

YingXue:
Update guide text for hub permission selection page
Fix accessiblity bug for filter box color contrast

### General Guidelines

### DIFF
--- a/src/app/authentication/components/authenticationSelection.scss
+++ b/src/app/authentication/components/authenticationSelection.scss
@@ -4,7 +4,7 @@
 **********************************************************/
 .auth-slection-container {
     display: flex;
-    padding-top: 45px;
+    margin: 45px;
     justify-content: center;
     .auth-selection-tile {
         padding: 50px;
--- a/src/localization/locales/en.json
+++ b/src/localization/locales/en.json
@@ -975,9 +975,9 @@
         }
     },
     "authentication": {
-        "autheSelection": {
+        "authSelection": {
             "header": "Welcome to Azure IoT Explorer, a cross-platform UI for interacting with devices attached to Azure IoT Hub",
-            "subText": "Choose an authentication method and connect to an Azure IoT hub",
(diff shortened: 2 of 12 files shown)


src/app/authentication/components/authenticationSelection.scss
```diff
@@ -4,7 +4,7 @@
 **********************************************************/
 .auth-slection-container {
     display: flex;
-    padding-top: 45px;
+    margin: 45px;
```
c-ryan-k, line 7: Just to make sure - did this need to be `margin-top` or was the guidance / fix to make a margin around the entire container?

YingXue: This is not directly related to the accessibility bug, but something I've noticed while working on this page. There is no margin which made the component look weird when the screen is small.

src/localization/locales/en.json
```diff
@@ -975,9 +975,9 @@
         }
     },
     "authentication": {
-        "autheSelection": {
+        "authSelection": {
             "header": "Welcome to Azure IoT Explorer, a cross-platform UI for interacting with devices attached to Azure IoT Hub",
-            "subText": "Choose an authentication method and connect to an Azure IoT hub",
+            "subText": "Choose an authentication method and connect to an Azure IoT hub. IoT Hub uses permissions to grant access to each IoT hub endpoint. Permissions limit the access to an IoT hub based on functionality. <0>Learn more</0>",
```
c-ryan-k, line 980: nit - usage of `IoT Hub` vs `IoT hub` - Do they all need capitalized?

c-ryan-k (approved):
Code changes look good, just had a couple of nits that I don't think are blockers

Elsie4ever (approved):

src/localization/locales/en.json
YingXue: I copied this sentence from the official doc. I think if it's an instance it will be lower case, and if not we use capitialized letters :)

PR #737, opened 2026-07-29
Accessibility fixes

raharri:
# Changes
- modify colors and styling in dark mode to match color contrast requirements
- announce changes in the shared copy button popover text
- make aria-labels for copy and query parameter elements more descriptive
- return focus to the add connections button after closing the 'Add connection string' dialog

---
This project has adopted the [Microsoft Open Source Code of Conduct](https://opensource.microsoft.com/codeofconduct/). For more information see the [Code of Conduct FAQ](https://opensource.microsoft.com/codeofconduct/faq/) or contact [opencode@microsoft.com](mailto:opencode@microsoft.com) with any additional questions or comments.

Thank you for contributing to the Azure IoT Explorer!

This checklist is used to make sure that common guidelines for a pull request are followed.

### General Guidelines


vincenzocaruso (approved):

vincenzocaruso (dismissed):

src/app/css/_themes.scss
```diff
@@ -106,7 +106,7 @@
     inputComponentBackgroundColor: $dark-gray,
     pivotOnHoverBackground: #363636,
     menuLinkBackground_Active: rgb(37, 37, 37),
-    menuLinkBackground_Hover: rgb(234, 234, 234),
+    menuLinkBackground_Hover: $dark-medium-gray,
     menuLinkBorder_Active: rgb(75, 166, 216),
     menuLinkColor_Hover: $link-blue,
```
rafarhat: **Blocking — this regresses contrast to 1.68:1.**

Both `homeViewNavigation.scss` and `deviceContentNav.scss` set `color` and `background-color` together in the same `a:hover` block:

```scss
a:hover {
    @include themify($themes) {
        color: themed('menuLinkColor_Hover');
        background-color: themed('menuLinkBackground_Hover');
    }
}
```

So moving the hover background to `$dark-medium-gray` (`rgb(49,51,53)`) while `menuLinkColor_Hover` stays `$link-blue` (`#0051ad`) gives dark blue on dark gray:

| | background | ratio |
|---|---|---|
| before | `rgb(234,234,234)` | 6.28:1 ✅ |
| after | `rgb(49,51,53)` | **1.68:1** ❌ |

Killing the white flash in dark mode is the right call, but the text has to come with it. Suggest also setting in the dark map:

```scss
menuLinkColor_Hover: rgb(75, 166, 216),
```

That's **4.68:1** on the new background, and it's already the value used for `linkColor` and `menuLinkBorder_Active` in this same map, so it stays consistent.

src/app/constants/fluentV9Theme.ts
```diff
@@ -51,7 +51,16 @@ const darkBrand: BrandVariants = {
...
-export const v9ThemeDark: Theme = createDarkTheme(darkBrand);
+
+// Override the dark theme's primary button colors to maintain sufficient contrast with white text.
+// Uses primary button colors from Azure portal dark mode
+export const v9ThemeDark: Theme = {
+    ...createDarkTheme(darkBrand),
+    colorBrandBackground: 'rgb(0, 120, 212)',
+    colorBrandBackgroundHover: 'rgb(16, 110, 190)',
+    colorBrandBackgroundPressed: darkBrand[20],
+    colorBrandBackgroundSelected: darkBrand[30],
```
rafarhat, line 62: The two colors you did change check out: `#0078D4` on white is **4.53:1** and `#106EBE` is **5.26:1** (vs. **2.71:1** for the `darkBrand[70]` the theme was deriving before). Nice catch.

Two follow-ups though:

**1. The sibling brand tokens still fail.** `createDarkTheme(darkBrand)` also derives `colorCompoundBrandBackground` / `...Hover` / `...Pressed` and `colorBrandBackgroundStatic` from `darkBrand[70]` = `#4ba6d8`. Those drive Checkbox, Radio, Switch and Slider fills, which render a white glyph on top — still **2.71:1**. So checked checkboxes/switches in dark mode are non-compliant even after this fix. Either shift the ramp itself (set `darkBrand[70]`/`[80]` to `#0078D4`/`#106EBE`, which fixes every derived token at once) or override the compound tokens explicitly here.

**2. Pressed/Selected don't match the stated source.** The comment says these come from Azure portal dark mode, but `darkBrand[20]` = `#0c1b23` and `darkBrand[30]` = `#173241` are near-black — a jarring jump from a `#0078D4` rest state and `#106EBE` hover. Portal uses `#005A9E` for pressed. Something like:

```ts
colorBrandBackgroundPressed: 'rgb(0, 90, 158)',   // #005A9E
colorBrandBackgroundSelected: 'rgb(16, 110, 190)', // #106EBE
```

keeps the ramp coherent and still gives 7:1+ against white.

Also a small robustness point: `#0078D4` clears AA by 0.03. If anyone ever nudges the button's text opacity or adds a translucent overlay it drops below 4.5:1. `#0067B8` buys margin if you want it.

src/app/notifications/components/notificationListEntry.tsx
```diff
@@ -44,8 +44,8 @@ export const NotificationListEntry: React.FC<NotificationListEntryProps> = (prop
             <Button
                 appearance="subtle"
                 icon={<CopyRegular />}
-                title={t(ResourceKeys.header.notifications.copy)}
-                aria-label={t(ResourceKeys.header.notifications.copy)}
+                title={t(ResourceKeys.header.notifications.copy.label)}
+                aria-label={t(ResourceKeys.header.notifications.copy.ariaLabel, { notification: message })}
```
rafarhat: Two issues with interpolating the raw message here.

**1. i18next escapes interpolated values by default.** Notification bodies routinely contain `'`, `"`, `&` and `<` (device IDs, connection strings, error text), so a message like `Device 'foo' failed` becomes the literal string `Copy notification: Device &#39;foo&#39; failed`, and that's what a screen reader announces — entity codes and all. Needs:

```tsx
aria-label={t(ResourceKeys.header.notifications.copy.ariaLabel, {
    notification: message,
    interpolation: { escapeValue: false }
})}
```

(No XSS concern — it's an attribute value, not `dangerouslySetInnerHTML`.)

**2. Length.** These messages can be several sentences (or a serialized error). An accessible name that long gets re-read on every arrow/tab pass through the notification list and is genuinely painful to navigate. Consider interpolating the notification *title* instead of the body, or truncating to ~80 chars — the full text is already available in the list item itself.

src/app/connectionStrings/components/connectionStringsView.tsx
```diff
@@ -64,12 +65,20 @@ export const ConnectionStringsView: React.FC = () => {
...
+        }
     };
 
     const onConnectionStringEditDismiss = () => {
+        const wasAdding = connectionStringUnderEdit === '';
         setConnectionStringUnderEdit(undefined);
+        if (wasAdding) {
+            addButtonRef.current?.focus();
+        }
     };
```
rafarhat, line 87: The bug being fixed is real (WCAG 2.4.3 — focus falls to `<body>` on dismiss because `ConnectionStringEditView` is rendered conditionally from state rather than via a `<DrawerTrigger>`, so Fluent has no trigger to restore to). Three problems with this implementation:

**1. It races the focus trap.** `ConnectionStringEditView` is an `OverlayDrawer`, i.e. a modal with a Tabster modalizer. `addButtonRef.current?.focus()` runs synchronously in the same handler as `setConnectionStringUnderEdit(undefined)`, so it fires **while the drawer is still mounted and still trapping focus** — the trap can pull focus straight back. This will pass in jsdom and be flaky in a real browser.

**2. It only covers the add path.** `wasAdding` gates everything, so dismissing the *edit* drawer still drops focus to `<body>` — same bug, different entry point.

**3. It misses Esc / overlay-click.** Those route through `onOpenChange` → `onDismiss`, so they happen to hit this handler, but the pattern is fragile.

All three fall out if you use the v9 primitives instead of an imperative `.focus()`:

```tsx
// in ConnectionStringsView (or on each trigger)
const restoreFocusTargetAttrs = useRestoreFocusTarget();

// in ConnectionStringEditView
const restoreFocusSourceAttrs = useRestoreFocusSource();
<OverlayDrawer {...restoreFocusSourceAttrs} ... />
```

Fluent then restores focus to whichever element opened the drawer — add *or* edit — after the trap releases. If you'd rather keep the ref approach, at minimum move the `.focus()` into a `React.useEffect` keyed on `connectionStringUnderEdit` transitioning to `undefined` so it runs after unmount.

One more: in `onConnectionStringEditCommit`, a successful add triggers the `synchronizationStatus === upserted` effect which navigates to the device list — so focusing the add button there is a no-op at best, and the destination route should be owning focus instead.

src/app/connectionStrings/components/connectionStringsView.tsx
```diff
@@ -29,6 +29,7 @@ export const ConnectionStringsView: React.FC = () => {
     useBreadcrumbEntry({name: t(ResourceKeys.breadcrumb.resources)});
     const [ state, api ] = useConnectionStringContext();
     const [ connectionStringUnderEdit, setConnectionStringUnderEdit ] = React.useState<string>(undefined);
+    const addButtonRef = React.useRef<HTMLButtonElement>(null);
```
rafarhat, line 32: The PR checklist says new behavior is backed by unit tests, but the focus-restore change — arguably the highest-risk one here — doesn't have one. Worth adding to `connectionStringsView.spec.tsx`: open the drawer, dismiss it, assert `document.activeElement` is the add button. Same for the edit path once that's covered.

src/app/shared/components/copyButton.tsx
```diff
@@ -21,6 +22,14 @@ export const CopyButton: React.FC<CopyButtonProps> = ({ copyText, disabled }) =>
         setCalloutTextKey(ResourceKeys.common.maskedCopyableTextField.copy.label);
     };
 
+    const handlePopoverOpenChange = (_event: unknown, data: { open: boolean }) => {
+        if (data.open) {
+            setPopoverOpen(true);
+        } else {
+            dismissPopover();
+        }
+    };
```
rafarhat: Centralizing the reset here is a real improvement — it fixes the stale "Copied" text on Esc and outside-click, not just mouse-leave. Three notes:

**1. A second click now copies with no visible feedback.** `PopoverTrigger` toggles, so click #2 fires `copyToClipboard` (sets `copied`, announces "Copied") *and* `onOpenChange(open: false)` → `dismissPopover()` (resets to `copy`). Both are in the same React event, so the batched result is a closed popover showing nothing. Screen reader users get the announcement; sighted users get silence on a repeat copy. Either keep the popover open on click, or drive `open` purely from hover/focus and let `copyToClipboard` force it open.

**2. Still no `onBlur`.** `onFocus` opens the popover but nothing closes it on blur, so a keyboard user who tabs away leaves it hanging on screen. Pre-existing, but this is the natural place to add `onBlur={dismissPopover}`.

**3. Types.** `(_event: unknown, data: { open: boolean })` hand-rolls a signature Fluent already exports. `PopoverProps['onOpenChange']` (or `OnOpenChangeData`) will stay correct if the signature changes upstream.

Broader thought for a follow-up, not this PR: this is a tooltip implemented with `Popover`, so the surface isn't exposed as `role="tooltip"`. `<Tooltip relationship="label">` is the idiomatic v9 control here — the `announce()` you added is the right escape hatch given the current structure, but the structure itself is the root cause.

src/app/css/_header.scss
```diff
@@ -60,6 +60,9 @@
         .settings button {
             color: themed('headerButtonColor');
         }
+        .settings button:focus-visible {
+            outline-color: themed('headerButtonColor');
+        }
```
rafarhat: Worth double-checking this actually renders — I suspect it's a no-op.

The `.settings` buttons are Fluent v9 `Button`s (`settingsPane.tsx`), and v9 draws its focus indicator with an `::after` pseudo-element bordered by `--colorStrokeFocus1`/`--colorStrokeFocus2`, not with `outline`. On top of that, where the UA outline is `outline-style: auto` (Chromium's default), `outline-color` is ignored entirely.

If the goal is a visible focus ring against the dark header, overriding the token on this scope is the durable fix:

```scss
.settings button { --colorStrokeFocus2: #{themed('headerButtonColor')}; }
```

Either way, please confirm visually in dark mode + keyboard nav before merging.

src/app/css/_notification.scss
```diff
@@ -84,6 +84,9 @@
 
 .notification-toast-body {
     word-break: break-word;
+    @include themify($themes) {
+        background-color: themed('notificationBackgroundColor') !important;
```
rafarhat: The `!important` here is a smell worth resolving rather than shipping. `.toast-notification` (line 94) already themes the toast root with the same token, so needing a second `!important`-ed background on the *body* element suggests a specificity fight with react-toastify v11's own `--toastify-color-*` variables.

Cleaner: override those vars on the container (or pass `theme` to `<ToastContainer>`) instead of out-`!important`-ing the library. While you're here, please confirm whether the `.toast-notification` rule is now dead — if it is, drop it so the next person isn't chasing two sources of truth.

Separately: `notificationBackgroundColor: var(--colorNeutralBackground1)` in `_themes.scss` works because `FluentProvider` sits above everything, but if that var is ever undefined the whole declaration is dropped and the toast renders transparent. A hardcoded fallback — `var(--colorNeutralBackground1, #292929)` — makes that failure mode non-catastrophic.

src/localization/locales/en.json
```diff
@@ -417,7 +420,7 @@
                     },
                     "value": {
                         "placeholder": "Select value",
-                        "ariaLabel": "Enter a value for this clause",
+                        "ariaLabel": "Query clause value",
```
rafarhat: Replacing the placeholder-as-label is the right fix (a placeholder disappears on input and isn't a reliable accessible name), but the new string is *less* descriptive than the one it replaces — "Enter a value for this clause" → "Query clause value".

More importantly, with multiple query pills on screen every clause dropdown now has an identical accessible name, so a screen reader user tabbing through hears "Query clause value" three times with no way to tell which clause they're in. Since the parameter type is already selected by then, something like `Value for {{parameter}}` (interpolating the selected parameter type) would disambiguate them.

src/app/shared/components/commandBarV9.tsx
```diff
@@ -77,6 +78,7 @@ export const CommandBarV9: React.FC<CommandBarV9Props> = ({ items, farItems, cla
         return (
             <ToolbarButton
                 key={item.key}
+                ref={item.buttonRef}
```
rafarhat, line 81: Correctly wired (`ToolbarButton` forwards its ref, and this only touches the non-submenu branch) — no bug here, just a design note.

`buttonRef` is a single-purpose escape hatch on a generic component. If the focus-restore switches to `useRestoreFocusTarget()` as suggested on `connectionStringsView.tsx`, you'd need to spread *attributes* onto the generated button rather than a ref, and you'd be back here adding a second prop. A generic `buttonProps?: Record<string, unknown>` (or `useImperativeHandle` on `CommandBarV9` exposing `focusItem(key)`) covers both cases and anything similar that comes up next.

src/app/css/_themes.scss
raharri: I'll dig into this bug but for some reason it's actually irrelevant because the menuLinkColor doesn't seem to affect the actual menu items either way, the color is being set somewhere else. So the solution might be just deleting these variables so they stop confusing the LLMs. The joy of CSS!

src/localization/locales/en.json
raharri: made the aria-label more specific

src/app/css/_notification.scss
raharri: removed the important and added a fallback

src/app/connectionStrings/components/connectionStringsView.tsx
raharri: useRestoreFocusSource doesn't quite behave the way this comment suggests it does, but used onDismiss() and made the implementation more generic to handle the edit scenario as well

src/app/shared/components/commandBarV9.tsx
raharri: see comment on connectionStringsView

src/app/constants/fluentV9Theme.ts
raharri: In practice, I don't see that in dark mode. Here's Claude's comment:
I need to verify the reviewer's comment on the fluentV9Theme token—they claimed `colorCompoundBrandBackground` should use `darkBrand[70]`, but that's incorrect. The actual value is `brand[100]` which equals `#95ccea`, and I'm checking the luminance calculation for that color against white glyphs to confirm the contrast.
I'm calculating the contrast ratio for the checked checkbox in dark mode, where the checkmark uses a dark grey (#242424) on the light blue background (#95ccea). Working through the luminance calculations, I get a ratio of about 8.95, which exceeds the WCAG AA standard.

vincenzocaruso (approved):

